PR #142, opened 2015-01-05
set_tcp_* functions should be static 

cuinutanix:
since they are not used anywhere else.

set_tcp_sockopt in particular conflicts with libnfs's function of the same
name and prevents a a program from statically linking against both libnfs and
libiscsi.

Similar fix should also go into libnfs.


sahlberg:
Merged both patches. Thanks.

PR #144, opened 2015-02-23
iscsi-command: fix unsolicited data-out length

plieven:
the recent implementation allows to send
iscsi->first_burst_length + iscsi->target_max_recv_data_segment_length
bytes if immediate and unsolicited data-out is send and
iscsi->target_max_recv_data_segment_length < iscsi->first_burst_length.

RFC3720 defines the length as:

Length=(min(FirstBurstLength, Expected Data
   Transfer Length) - Received Immediate Data Length).

so that immediate data and unsolicited data-out are together
FirstBurstLength at maximum.

### DIFF
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -33,6 +33,13 @@ extern "C" {
 #define discard_const(ptr) ((void *)((intptr_t)(ptr)))
 #endif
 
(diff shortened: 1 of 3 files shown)


sahlberg:
merged.  thanks

PR #145, opened 2015-03-01
Iscsi perf clang

sitsofe:
When trying to compile libiscsi on OS X 10.9 I hit a few clang warnings and the new iscsi-perf makes use of clock_gettime which doesn't existing on OS X (see http://nadeausoftware.com/articles/2012/03/c_c_tip_how_measure_cpu_time_benchmarking#Discussion and http://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x ). These patches fix these up (and tidy up some spacing in the util/Makefile.am while I'm there).

Looking at the fio source (https://github.com/axboe/fio/blob/1f440ece6127d2d274b961b6af0aa9091787702a/engines/posixaio.c#L36 ) shows that project uses gettimeofday when clock_gettime is unavailable so I do something similar here with autoconf (-lrt is always added to LIBS if we have clock_gettime) and ifdef headers. gettimeofday is bad because it's not necessarily monotonic and has usec accuracy at best but it's that or Mac OS X only ifdefs and code...

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -128,6 +128,9 @@ AC_CHECK_MEMBER([struct CU_SuiteInfo.pSetUpFunc],
 #include <CUnit/TestDB.h>
 ]])
 
+AC_SEARCH_LIBS(clock_gettime, rt, [
+	       AC_DEFINE([HAVE_CLOCK_GETTIME],1,[Define if clock_gettime is available])])
+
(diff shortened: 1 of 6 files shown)


sahlberg:
Merged, thanks!

PR #161, opened 2015-04-24
Fix '0block' write same and unmap tests and 'out-of-tree' builds

czankel:
This patchset fixes writesame and unmap tests for targets that have WSNZ bit enabled by skipping the affected tests (printing a proper message) including a couple of other fixes where the number of blocks passed to writesame and unmap was incorrect. It also fixes 'out-of-tree' builds and adds two generated files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,8 @@ TAGS
 /test-tool/Makefile.in
 /test-tool/Makefile
 /test-tool/iscsi-test-cu
+/tests/Makefile.in
+/tests/Makefile
 /utils/Makefile.in
 /utils/Makefile
 /utils/iscsi-inq
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -1,4 +1,4 @@
-AM_CPPFLAGS=-I. -I../include "-D_U_=__attribute__((unused))" \
+AM_CPPFLAGS=-I. -I${srcdir}/../include "-D_U_=__attribute__((unused))" \
 	"-D_R_(A,B)=__attribute__((format(printf,A,B)))"
 AM_CFLAGS=$(WARN_CFLAGS)
 LDADD = ../lib/libiscsi.la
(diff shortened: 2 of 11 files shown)


sahlberg:
Merged, thanks!

sitsofe:
Hmm I don't know that it is correct to skip the test entirely even though WSNZ is not supported. What should happen is that the operation should fail with a particular sense code and if the sense coming back is wrong that that in and of itself is an error...

czankel:
Hi Sitsofe,


On Sun, Apr 26, 2015 at 10:15 AM, Sitsofe Wheeler <notifications@github.com>
 wrote:

> Hmm I don't know that it is correct to skip the test entirely even though
> WSNZ is not supported. What should happen is that the operation should fail
> with a particular sense code and if the sense coming back is wrong that
> that in and of itself is an error...
>

The patch included the following check to make sure it would fail with the
correct sense code. I only added this check to two of the four tests,
though, so can provide a patch for the other two. Do you think that such a
test is not sufficient?


if (inq_bl->wsnz) {
ret = writesame10(sd, 0,
block_size, 0, 0, 0, 0, 0, buf,
EXPECT_INVALID_FIELD_IN_CDB);
CU_ASSERT_EQUAL(ret, 0);

Thanks,
-Chris

sahlberg:
You mean this patch:
commit fff67ba7dd5978ae12d068b807e9d69d300eb761
Author: Chris Zankel <chris@zankel.net>
Date:   Wed Apr 22 18:05:13 2015 -0700

    test_writesame*_0blocks: skip 0-block test if unsupported


I think it is fine as is.

We check that wsnz==1 does return proper sense codes once for each of the
two opcodes in the _0blocks.c tests.


I don't think it needs to test for this condition in  _unmap_until_end.c
also and that it should be fine to just skip the test there.

Of course, it does not hurt anything to have the test there too, so that a
target with broken wsnz would fail
all four instead of failing two and skipping two but i don't think it is
needed.


If you really want the test for proper sense code also in the other two
tests too then the code can be added.
I am not against it, I just don't think it is needed.

PR #166, opened 2015-05-28
test: multipath support

ddiss:
- Add support to iscsi-test-cu for multiple iscsi:// URLs. When multiple are provided, connect to each and check that they correspond to the same logical unit.
- Add a simple multipath IO test that checks for consistent data across multiple paths.


ddiss:
A tests_multipathio entry should also be added to the all_suites and scsi_suites arrays. I'll do that and resubmit.

PR #167, opened 2015-05-28
test: multipath support v2

ddiss:
- Add support to iscsi-test-cu for multiple iscsi:// URLs. When multiple are provided, connect to each and check that they correspond to the same logical unit.
- Add a simple multipath IO test that checks for consistent data across multiple paths.

Changes since v1:
- Add tests_multipathio entry to all_suites and scsi_suites.

### DIFF
--- a/test-tool/Makefile.am
+++ b/test-tool/Makefile.am
@@ -8,6 +8,7 @@ EXTRA_DIST = README
 
 dist_noinst_HEADERS = iscsi-support.h \
 		      iscsi-test-cu.h
+		      iscsi-multipath.h
 
 # libiscsi test tool using cunit
 if ISCSITEST
@@ -15,6 +16,7 @@ bin_PROGRAMS = iscsi-test-cu
 iscsi_test_cu_LDFLAGS = -ldl -lcunit
 iscsi_test_cu_SOURCES = iscsi-test-cu.c \
 	iscsi-support.c \
+	iscsi-multipath.c \
 	test_compareandwrite_simple.c \
 	test_compareandwrite_dpofua.c \
 	test_compareandwrite_miscompare.c \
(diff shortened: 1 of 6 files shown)


ddiss:
Ping, anyone get a chance to look at this?
I can send the patches to the mailing list if preferred.

sahlberg:
Merged, thanks.

I am not sure this test should be in all the families, but the families are
a bit messy right now.


I think at a later point I would like to divide up the tests as:

SCSI :   only tests that relate to scsi opcodes
iSCSI : tests for the iscsi protocol, which is where this multipath test
would be)

LINUX:  which is where we can have whatever tests for linux
interoperability.
This family currently contains heaps of SCSI tests, but should probably be
very different.

The idea I had behind the LINUX family was that this would be a family of
tests where we could add tests
that are beyond what the actual standard itself requires, and be sort of a
compliance test suite for devices that
want to interoperate well with linux.

These tests would thus be for things way beyond what the standards require.
But would be a set of tests for how you want a device to behave.
For example, WRITESAME16 is not mandatory, but if Linux wants a test to
suggest to vendors "ssd devices should wupport WS16 to work optimally with
linux"
we could then have a test that would "if ssd but WS16 is not implemented
then FAIL test".

This way linux could have a set of tests to provide to vendors as a
suggestion to what features that linux kernel wants and codify desired
behaviour
for SCSI devices in a test tool that can be dsitributed to them.



regards
ronnie sahlberg

ddiss:
Thanks for the explanation Ronnie - I'll keep this in mind in future.
I plan on adding a few more multipath specific tests for things like LU reservations, ALUA, COMPARE AND WRITE, etc. I'll ensure that they're added to the iSCSI family only.

sahlberg:
Well,   if we make the multipathing support /dev/* nodes as well,
so you can use multipathing on SGIO devices, then we can have it in the SCSI family.

PR #171, opened 2015-06-03
test_modesense6_control_swp: skip test if target cannot change swp

rodrigc:
Signed-off-by: Vadim Finkelstein <vadim@pi-coral.com>


sahlberg:
Merged, thanks

PR #174, opened 2015-06-12
test_modesense6_control_swp test: read current control before modesel…

agrare:
…ect6

After checking if SWP can be changed the current control page has
to be read back before sending it to initiator with modeselect6.


sahlberg:
merged, thanks

PR #180, opened 2015-08-27
Allow for autogen.sh failure

rcsheets:
I noticed that if autoconf isn't installed, autogen.sh printed a confusing combination of an error message and instructions to continue, and exited successfully even though it had failed:

```
$ ./autogen.sh 
./autogen.sh: 7: ./autogen.sh: autoreconf: not found
Now run ./configure and then make.
$ echo $?
0
```

This small change to the autogen.sh script allows it to check that the commands it needs are available, and if necessary to report which commands are missing, and exit unsuccessfully.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,5 +1,22 @@
 #!/bin/sh
 
+set -e
+
+needed='rm mkdir autoreconf echo'
+if ! type $needed >/dev/null
+then
+  for cmd in $needed
+  do
+    if ! type $cmd >/dev/null
+    then
+      # Have type print an error message for each missing command
+      type $cmd || true
+    fi
+  done
+  echo A required command is missing. Unable to continue.
+  exit 1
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -1,5 +1,23 @@
 #!/bin/sh
 
+set -e
+
+# Commands this script needs
+needed='rm mkdir autoreconf echo'
+if ! type $needed >/dev/null
```
lucasrangit: `type` will output to stderr when not found, so use `if ! type $needed &>/dev/null` here and on line 11.

rcsheets: That's what I expected it to do too, but not what I observed on my system. All the output went to `stderr`. Still, it's probably safer to redirect both output streams. I'll update the PR.

sahlberg:
Merged, Thanks!

PR #184, opened 2015-09-30
Persistent Reservation REPORT CAPS and PREEMPT tests

ddiss:
PREEMPT is currently only used to remove a registration - more tests to follow.


ddiss:
@sahlberg any thoughts on these changes?

Cheers, David

PR #190, opened 2016-02-23
TESTS: add 4MB overlap test to write10_simple

ddiss:
Ceph RADOS block devices split images into 4MB objects. This test
confirms that cross object IOs are successful.

Signed-off-by: Daniel Oliveira <doliveira@suse.com>


ddiss:
Thanks Ronnie - closing.

PR #191, opened 2016-03-10
Tests: Asynchronous iSCSI

ddiss:
Dispatch a number of READ10 requests simultaneously, and await all
responses.

The iscsi-support test helper functions are currently all synchronous,
so this test mostly uses the bare libiscsi API. As new async tests are
added, we can make a more informed decision on which boilerplate code
can be split out.


sahlberg:
Merged, thanks!

On Thu, Mar 10, 2016 at 10:45 AM, David Disseldorp <notifications@github.com

PR #193, opened 2016-03-15
iscsi: add async abort task test

ddiss:
These values are defined in rfc3720
10.6. Task Management Function Response
-> 10.6.1.  Response

The response field is a single byte value, and is already used within
libiscsi as a command_data parameter for iscsi_task_mgmt_async()
callbacks.

### DIFF
--- a/include/iscsi.h
+++ b/include/iscsi.h
@@ -595,6 +595,17 @@ enum iscsi_task_mgmt_funcs {
      ISCSI_TM_TASK_REASSIGN     = 0x08
 };
 
+enum iscsi_task_mgmt_response {
+	ISCSI_TMR_FUNC_COMPLETE				= 0x0,
+	ISCSI_TMR_TASK_DOES_NOT_EXIST			= 0x1,
+	ISCSI_TMR_LUN_DOES_NOT_EXIST			= 0x2,
+	ISCSI_TMR_TASK_STILL_ALLEGIANT			= 0x3,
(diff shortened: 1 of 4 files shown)


test-tool/test_async_abort.c
```diff
@@ -0,0 +1,251 @@
...
+		scsi_free_scsi_task(wr_tasks[i]);
+	}
+
+	/* any writes that were successful should have failed aborts */
+	CU_ASSERT_EQUAL(state.wr_good, state.abort_bad_itt);
+
+	/*
+	 * any successful aborts shouldn't have corresponding write completions
+	 */
+	CU_ASSERT_EQUAL(state.abort_ok, state.wr_dispatched - state.wr_good);
```
ddiss: Actually, these assertions might not quite match rfc3720 -
6.9.  SCSI Timeouts
...
   If the abort request is received and the original command is missing,
   targets MUST consider the original command with that RefCmdSN to be
   received and issue a Task Management response with the response code:
   "Function Complete".  This response concludes the task on both ends.
   If the abort request is received and the target can determine (based
   on the Referenced Task Tag) that the command was received and
   executed and also that the response was sent prior to the abort, then
   the target MUST respond with the response code of "Task Does Not
   Exist".

So not all state.abort_ok responses imply that a task has been aborted at the target - it may also mean that the abort was handled without knowledge that RefCmdSN was handled previously.

I'll withdraw this test for now, and rework it to track all WRITE10 *and* TMF requests individually.

ddiss:
Superseded.

PR #194, opened 2016-03-17
Fix Qemu Segfault

plieven:
Hi Ronnie,

I found Qemu segfaulting on storage timeouts with recent libiscsi. The reason you can find in the commits. I think we should invoke all callbacks of PDUs we drop, shouldn't we?

Peter


sahlberg:
thanks

PR #195, opened 2016-03-22
Remove local cancellation from ABORT TASK path and add simple test

ddiss:


sitsofe:
A variant of test_get_clock_sec already exists in https://github.com/sahlberg/libiscsi/blob/master/utils/iscsi-perf.c . Perhaps the two could be unified somewhere?

ddiss:
Yeah, I saw that but figured that it's small enough not to bother. I can move it somewhere common if you'd prefer - iscsi.h seems a little to public... any suggestions?

sitsofe:
How about iscsi-private.h ? Failing that perhaps there needs to be an iscsi-util.h to collect this style of function that could be useful across test tools and utility tools...

sahlberg:
Merged. thanks

Issue #189, opened 2016-02-03
Can't get iscsi-inq tool to properly log-in with CHAP or Mutual CHAP


bramklg:
I'm trying to use `iscsi-inq` tool to verify if specified iscsi settings are correct. This works for targets that don't require authentication but fails for targets that require CHAP or Mutual CHAP. I was first using version 1.4.0-3 that comes with ubuntu 14.04 but currently i'm testing against the master branch.

My ISCSI target is a freenas server that displays the following in the logs when trying to log-in:

> Feb  3 09:50:14 freenas ctld[94005]: 192.168.8.179 (iqn.2007-10.com.github:sahlberg:libiscsi:iscsi-inq): CHAP authentication failed for user "bram"
> Feb  3 09:50:14 freenas ctld[92923]: child process 94005 terminated with exit status 1

I'm certain the username and password are correct. When I set-up an iscsi connection using `iscsiadm` everything works as expected. Am I doing something wrong or is this a bug in the tool? 

This is the configuration of my iscsi target:
>host: 192.168.8.106
>target: iqn.2005-10.org.freenas.ctl:target2
>lun: 0
>Auth: CHAP
>User: bram
>Pass: test123456789

I'm using the following command to test if the connection settings are valid or not:
`iscsi-inq --debug iscsi://bram%test123456789@192.168.8.106:3260/iqn.2005-10.org.freenas.ctl:target2/0`

Executing this command gives the following output:
>libiscsi:1 Failed to log in to target. Status: Authentication failure(513) [iqn.2005-10.org.freenas.ctl:target2]
>Login Failed. Failed to log in to target. Status: Authentication failure(513)

If I execute the tool to a target without authentication requirements iscsi-inq correctly displays information about the target / connection.

sitsofe:
I think I've seen this too and the problem might have linked to if the target returned CHAP responses in base64 or binary (but I could be wrong).

sahlberg:
On Wed, Feb 3, 2016 at 1:01 AM, bramklg <notifications@github.com> wrote:

> I'm trying to use iscsi-inq tool to verify if specified iscsi settings
> are correct This works for targets that don't require authentication but
> fails for targets that require CHAP or Mutual CHAP I was first using
> version 140-3 that comes with ubuntu 1404 but currently i'm testing against
> the master branch
>
> My ISCSI target is a freenas server that displays the following in the
> logs when trying to log-in:
>
[...]

PR #196, opened 2016-03-29
Fix 2 CHAP related issues

cardamon:
Hi!

This fixes two CHAP related issues that arise when using FreeNAS. The first is #189 .

#189 is caused by only reading and hashing the first 255 bytes of the CHAP challange.

From https://tools.ietf.org/html/rfc3720#section-11.1.4 :
> ...and C and R are large-binary-values and their binary length (not the length of the character string that represents them in encoded form) MUST not exceed 1024 bytes.

So a target may send a 2K hex string encoding 1K data...

The second issue is FreeNAS not being very lenient when sending too many headers; FreeNAS will immediately close the connection and log an error:
```
Mar 25 10:05:50 freenas ctld[98731]: 192.168.8.156 (iqn.2007-10.com.github:sahlberg:libiscsi:iscsi-inq): initiator resent InitiatorName
Mar 25 10:05:50 freenas ctld[1993]: child process 98731 terminated with exit status 1
```

I've ran `test_1000_chap.sh` and it still works, so `tgt` support is not affected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ TAGS
 /utils/iscsi-readcapacity16
 /utils/iscsi-swp
 /libiscsi.pc
+/.cproject
+/.project
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -63,6 +63,9 @@ void iscsi_free_iscsi_inqueue(struct iscsi_context *iscsi, struct iscsi_in_pdu *
 /* size of chap response field */
 #define CHAP_R_SIZE 16
 
+/* max length of chap challange */
+#define MAX_CHAP_C_LENGTH 2048
+
 struct iscsi_context {
 	char initiator_name[MAX_STRING_SIZE+1];
 	char target_name[MAX_STRING_SIZE+1];
@@ -74,7 +77,7 @@ struct iscsi_context {
 
 	char user[MAX_STRING_SIZE+1];
 	char passwd[MAX_STRING_SIZE+1];
-	char chap_c[MAX_STRING_SIZE+1];
+	char chap_c[MAX_CHAP_C_LENGTH+1];
 
 	char target_user[MAX_STRING_SIZE+1];
 	char target_passwd[MAX_STRING_SIZE+1];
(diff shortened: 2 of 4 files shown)


sahlberg:
I merged it manually after squashing the first three patches into a single patch.

Thanks!

mallilinux1980:
Hi,

May I know which version has this fixed?

Also much appreciated if anyone can give steps to send invalid cmdSN to iscsi target? Ideally as per iscsi RFC target should just ignore or reject I believe? Target should not become unresponsive.

sahlberg:
On Fri, Sep 27, 2019 at 8:53 AM mallilinux1980 <notifications@github.com>
wrote:

> Hi,
>
> May I know which version has this fixed?
>
This should be fixed in version 1.16 and later.

Also much appreciated if anyone can give steps to send invalid cmdSN to
> iscsi target? Ideally as per iscsi RFC target should just ignore or reject
> I believe? Target should not become unresponsive.
>
Look in the test-tool directory.
There are a bunch of tests in there that send invalid PDUs to the target,
such as cmd sequence numbers outside of the allowed window etc
See for example test_iscsi_cmdsn_toohigh.c  which is a test for what the
target does if the cmdsn is too high.  i.e. a test for invalid cmdsn

Also look at the recent commits in the master branch from David Disseldorp
which adds a few more tests like this.

sahlberg:
Oh,
and if you create new scsi/iscsi tests,   please add them to the test-suite
in libiscsi/test-tool instead of writing a new/different test tool.

mallilinux1980:
I tried by downloading a new version 1.19 from gitlab but still say
authorization issue.

The error screenshot is attached for reference. I've also tried target
discovery with given an initiator which is CHAP disabled without
credentials even then i get the same authorization issue.

I don't have problems whatsoever if I use plain iscsiadm commands.

mallilinux1980:
BTW, my OS is RHEL 6.7

mallilinux1980:
i've tried older version 1.7 using RPMs (libiscsi & libiscsi-utils) which
is part of RHEL6 distribution after I disable CHAP UN & PW forgiven a host
IQN (aka initiator) I find no authorization issues but test is aborted with
glibc memory backtrace. Attached screenshot for reference.

PR #197, opened 2016-03-30
Fix iscsi-ls for IPv6 link-local addresses, and bump ABI version again

bonzini:
Hi Ronnie, please take a look. Thanks!


sahlberg:
Merged. Thanks!

sahlberg:
This could be a useful test to add to the test-suite.

IF the test target is iscsi, do a discovery and check all portals that are returned.
Verify that neither v4/6 loopback nor v6 link local addresses are returned.

I will work on that test tomorrow,
Thanks again!

PR #199, opened 2016-03-31
Silence clang missing-field-initializers warning

sitsofe:
When libiscsi is built with clang several spurious
missing-field-initializers warnings like the following are reported:

test_async_write.c:61:45: error: missing field 'completed' initializer
[-Werror,-Wmissing-field-initializers]
	struct tests_async_write_state state = { 0 };

These appear to be due to clang being overzealous in its checking
because the code is legal C (see
https://llvm.org/bugs/show_bug.cgi?id=21689 ) but it causes compilation
to be aborted. Workaround the problem by specifying a 0 or NULL for each
structure member.

### DIFF
--- a/test-tool/test_async_abort_simple.c
+++ b/test-tool/test_async_abort_simple.c
@@ -87,7 +87,7 @@ void
 test_async_abort_simple(void)
 {
 	int ret;
-	struct tests_async_abort_state state = { 0 };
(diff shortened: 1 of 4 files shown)


sahlberg:
Meged.
memset() would be an alternative (which I prefer) but it is not worth it to change right now.

PR #203, opened 2016-04-05
Tests: Ensure each path sends a different initiator when using multipath

sitsofe:
Tests such as ProutPreempt presume each path they have is sending a
different initiator name. When using multipath this did not happen and
initiatorname1 was used on all paths.

Fix this by creating and initializing an array (currently size 2) of
initiator names and iterating it in step with each new path we make
(hopefully the compiler will complain if iscsi-test-cu ever supports
more than two multipaths and there isn't a corresponding initiatorname).

Signed-off-by: Sitsofe Wheeler <sitsofe@yahoo.com>

### DIFF
--- a/test-tool/iscsi-test-cu.c
+++ b/test-tool/iscsi-test-cu.c
@@ -765,6 +765,7 @@ int
 suite_init(void)
 {
         int i;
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged. Thanks

PR #213, opened 2016-07-25
test Compare And Write against unwritten range

ddiss:


sitsofe:
Perhaps this test should use `inq_bl->max_cmp` from block limits like https://github.com/sahlberg/libiscsi/blob/1.17.0/test-tool/test_compareandwrite_miscompare.c#L41 ?

ddiss:
@sitsofe ESXi, which this test is attempting to emulate, appears to only issue single block Compare And Write / ATS requests, so I'd prefer to keep the IO sizes as is.

Thanks for the feedback!

sitsofe:
@ddiss Is this test to catch the case where no one has ever written to that region of the target ever before (e.g the first time an initiator connects to a newly made thin provisioned target) or just the case where the initiator hasn't written to the target recently?

ddiss:
@sitsofe the former - first Compare And Write IO to an unwritten range on a new thin provisioned LU.

sitsofe:
I wonder if it would be good to additionally CAW blocks which are half way through the target... My reasoning is that there's a smaller chance of a prior test having already written (or read) from that region when running libiscsi-cu-test as a test suite...

sitsofe:
@ddiss Should there be a return; after the CU_PASS of line 56?

ddiss:
@sitsofe there should be - will push fix with some indentation cleanups.

ddiss:
@sitsofe did you get a chance to look at the latest version, and if so, do you have any feedback?
Regarding your suggestion to write half way through the LUN - I'd prefer to wait until trim/discard support is available with libiscsi, and then discard the entire region prior to the compare-and-write loop.

sahlberg:
I merged it. Thanks.

Libiscsi does have support for trim/discard.
Discard is available using the scsi calls for WRITESAME10/WRITESAME16/UNMAP opcodes

sitsofe:
@ddiss for what it's worth it looked fine to me (but I hadn't had a chance to run it). The only thing I would have said is that I think the halfway test would have been useful for thin devices that for whatever reason don't support trim...

ddiss:
Thanks!

sahlberg:
To test CAW onto unmapped blocks you can use GET_LBA_STATUS.
If the device supports thin-provisioning this command will return LBA
ranges and their status (unmapped/anchored/mapped).
That way you can discover if there are unmapped blocks and then explicitely
try COW onto them.

PR #214, opened 2016-08-05
abstract transport to static driver functions and opaque driver speci…

plieven:
@shterman please check if iSER still works. It compiles, but I have no chance to test it. Can you please also check if the paramters to ibv_create_cq in iser_addr_handler are correct.

### DIFF
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -76,9 +76,10 @@ union socket_address {
 };
 
 struct iscsi_context {
-
-	struct iscsi_transport *t;
+	struct iscsi_transport *drv;
+	void *opaque;
 	enum iscsi_transport_type transport;
+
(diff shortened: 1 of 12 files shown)


shterman:
@plieven I will do some sanity checks over the weekend. about different parameter in ibv_create_cq, it shouldn't matter because for now I'm not using this parameter which is (void*)cq_context.

plieven:
@shterman thank you,regarding the ibv_create_cq I was unsure if there is a callback somewhere which receives the parameter as an opaque void*. it would also be good to know if the checks for NULL in iscsi_iser_connect and iser_connected_handler are necessary or what was the intention to add them.

shterman:
@plieven  you are right, NULL checks in both functions are unnecessary.

shterman:
@plieven I did some sanity checks with your commit, iSER seems to work fine.

PR #217, opened 2016-10-05
TESTS: drop /dev/sg device path check for iSCSI

ddiss:
When run against an iSCSI target sd->sgio_dev may be NULL, so the
strncmp() coredumps.

Fixes: 698ae7a649e66a44f398cd35c9ac2118f7411487


sahlberg:
Thanks

PR #224, opened 2016-12-11
Batch pdu read in function iscsi_tcp_service()

Chilledheart:
iscsi_tcp_service() doesn't read all pdu available in the socket, while iscsi_tcp_service() writes all pdus in the outqueue. this patch changes this behavior and improves tcp performance.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -91,8 +91,8 @@ iscsi_add_to_outqueue(struct iscsi_context *iscsi, struct iscsi_pdu *pdu)
 		pdu->next = NULL;
 		return;
 	}
-	
-	/* queue pdus in ascending order of CmdSN. 
+
+	/* queue pdus in ascending order of CmdSN.
 	 * ensure that pakets with the same CmdSN are kept in FIFO order.
 	 * immediate PDUs are queued in front of queue with the CmdSN
 	 * of the first element in the outqueue.
@@ -117,7 +117,7 @@ iscsi_add_to_outqueue(struct iscsi_context *iscsi, struct iscsi_pdu *pdu)
 		last=current;
 		current=current->next;
 	} while (current != NULL);
-	
+
 	last->next = pdu;
 	pdu->next = NULL;
 }
@@ -562,9 +562,11 @@ iscsi_iovector_readv_writev(struct iscsi_context *iscsi, struct scsi_iovector *i
 	return n;
 }
 
-static int
-iscsi_read_from_socket(struct iscsi_context *iscsi)
-{
+/* read one pdu from socket
+ *
+ */
+static int iscsi_read_pdu_from_socket(struct iscsi_context *iscsi,
+									  struct iscsi_in_pdu **pdu) {
 	struct iscsi_in_pdu *in;
 	ssize_t data_size, count, padding_size;
 
@@ -586,6 +588,7 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 		count = ISCSI_HEADER_SIZE - in->hdr_pos;
 		count = recv(iscsi->fd, &in->hdr[in->hdr_pos], count, 0);
 		if (count == 0) {
+			iscsi_set_error(iscsi, "read from closed socket");
 			return -1;
 		}
 		if (count < 0) {
@@ -636,8 +639,9 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 			}
 			count = recv(iscsi->fd, buf, count, 0);
 		}
-		
+
 		if (count == 0) {
+			iscsi_set_error(iscsi, "read from closed socket");
 			return -1;
 		}
 		if (count < 0) {
@@ -653,6 +657,7 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 		in->data_pos += count;
 	}
 
+	/* not enough data */
 	if (in->data_pos < data_size) {
 		return 0;
 	}
@@ -660,7 +665,29 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 	ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
 	iscsi->incoming = NULL;
 
+	*pdu = in;
+	return 1;
+}
 
+// TODO(sahlberg) change this to 1 for AROS which doesn't support NONBLOCK io
+// actually
+#define READ_PDU_BATCH_COUNT 256
+
+static int iscsi_read_from_socket(struct iscsi_context *iscsi) {
+	int i;
+	/* we shouldn't do batch read on logging stage */
+	int batch_count  = iscsi->is_loggedin ? READ_PDU_BATCH_COUNT : 1;
+	for (i = 0; i < batch_count; ++i) {
+		struct iscsi_in_pdu *pdu;
+		int ret = iscsi_read_pdu_from_socket(iscsi, &pdu);
+		/* NODATA */
+		if (ret == 0) {
+			break;
+		}
+		if (ret < 0) {
+			return -1;
+		}
+	}
 	while (iscsi->inqueue != NULL) {
 		struct iscsi_in_pdu *current = iscsi->inqueue;
 
@@ -703,7 +730,7 @@ iscsi_write_to_socket(struct iscsi_context *iscsi)
 				ISCSI_LOG(iscsi, 6, "iscsi_write_to_socket: socket is corked");
 				return 0;
 			}
-			
+
 			if (iscsi_serial32_compare(iscsi->outqueue->cmdsn, iscsi->maxcmdsn) > 0
 				&& !(iscsi->outqueue->outdata.data[0] & ISCSI_PDU_IMMEDIATE)) {
 				/* stop sending for non-immediate PDUs. maxcmdsn is reached */
@@ -721,10 +748,10 @@ iscsi_write_to_socket(struct iscsi_context *iscsi)
 				return -1;
 			}
 			iscsi->outqueue_current = iscsi->outqueue;
-			
+
 			/* set exp statsn */
 			iscsi_pdu_set_expstatsn(iscsi->outqueue_current, iscsi->statsn + 1);
-			
+
 			ISCSI_LIST_REMOVE(&iscsi->outqueue, iscsi->outqueue_current);
 			if (!(iscsi->outqueue_current->flags & ISCSI_PDU_DELETE_WHEN_SENT)) {
 				/* we have to add the pdu to the waitqueue already here


plieven:
You are right that iscsi_read_from_socket behaves differently than iscsi_write_to_socket. Before merging this I would like to think about possible side implications this might have. For instance: can a constant amount of input packets lead to a denial of service scenario where we never exit iscsi_read_from_socket anymore. Furthermore might it be a good idea to call iscsi_write_to_socket after each PDU is received to decrease latency. And might it be better to do the batching inside iscsi_read_from_socket itself.

plieven:
I was thinking of sth like this. But I did not manage to receive more than one PDU so far...

`diff --git a/lib/socket.c b/lib/socket.c
index 036b4f4..5c33ebe 100644
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -563,10 +563,11 @@ iscsi_iovector_readv_writev(struct iscsi_context *iscsi, struct scsi_iovector *i
 }
 
 static int
-iscsi_read_from_socket(struct iscsi_context *iscsi)
+iscsi_read_from_socket(struct iscsi_context *iscsi, int *new_pdu_rcvd)
 {
        struct iscsi_in_pdu *in;
        ssize_t data_size, count, padding_size;
+       *new_pdu_rcvd = 0;
 
        if (iscsi->incoming == NULL) {
                iscsi->incoming = iscsi_szmalloc(iscsi, sizeof(struct iscsi_in_pdu));
@@ -659,7 +660,7 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 
        ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
        iscsi->incoming = NULL;
-
+       *new_pdu_rcvd = 1;
 
        while (iscsi->inqueue != NULL) {
                struct iscsi_in_pdu *current = iscsi->inqueue;
@@ -671,7 +672,6 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
                iscsi_free_iscsi_in_pdu(iscsi, current);
        }
 
-
        return 0;
 }
 
@@ -840,6 +840,8 @@ iscsi_service_reconnect_if_loggedin(struct iscsi_context *iscsi)
 static int
 iscsi_tcp_service(struct iscsi_context *iscsi, int revents)
 {
+       int new_pdu_rcvd = 0, num_pdus_rcvd = 0;
+
        if (iscsi->fd < 0) {
                return 0;
        }
@@ -925,17 +927,21 @@ iscsi_tcp_service(struct iscsi_context *iscsi, int revents)
                return 0;
        }
 
-       if (revents & POLLIN) {
-               if (iscsi_read_from_socket(iscsi) != 0) {
-                       return iscsi_service_reconnect_if_loggedin(iscsi);
+       do {
+               if (revents & POLLIN) {
+                       if (iscsi_read_from_socket(iscsi, &new_pdu_rcvd) != 0) {
+                               return iscsi_service_reconnect_if_loggedin(iscsi);
+                       }
                }
-       }
-       if (revents & POLLOUT) {
-               if (iscsi_write_to_socket(iscsi) != 0) {
-                       return iscsi_service_reconnect_if_loggedin(iscsi);
+               if (revents & POLLOUT) {
+                       if (iscsi_write_to_socket(iscsi) != 0) {
+                               return iscsi_service_reconnect_if_loggedin(iscsi);
+                       }
                }
-       }
-       iscsi_timeout_scan(iscsi);
+               iscsi_timeout_scan(iscsi);
+               num_pdus_rcvd += new_pdu_rcvd;
+       } while (new_pdu_rcvd && iscsi->is_connected && num_pdus_rcvd < 16);
+       if (num_pdus_rcvd > 1) ISCSI_LOG(iscsi, 2, "num pdus rcvd = %d\n", num_pdus_rcvd);
 
        return 0;
 }
`

sahlberg:
i like this proposal better but we still call read_from_socket() more than
we need, and we still also would break on platforms that lack non-blocking
sockets.
For the read case this would be real be real bad as it will cause the
application to hang indefinitely.


What I think we want to do here is only modify iscsi_tcp_service()
and do something like :


...
  if (revents & POLLIN) {
     while (1) {
           int count = 0;
  if (iscsi_read_from_socket(iscsi) != 0) {
  return iscsi_service_reconnect_if_loggedin(iscsi);
           }
           if (ioctl(... FIONREAD, &count) < 0 || count == 0) {
                return 0;
           }
  }
...

sahlberg:
I created a test for this in the tests subdirectory.
test_0410_read_all_pdus.sh

It currently fails as iscsi_service() still only consumes a single PDU per
time we invoke it.

Chilledheart:
Yes, it might be a problem that it starves for read events. In fact CPU goes much faster than network packet processing, so soon the read is drained or queue is full. After that, it returns. For the case of ioctl, it is good enough to detect the read event (I tried poll as well, and the performance degraded, i.e. 10%) but still might doesn't work with the platforms without non-block support.

Anyway I thought it is a standard routine (or at least common one) to detect read event by calling read() call to get an EAGAIN or EWOULDBLOCK. I see this trick in some popular event libraries such as libevent. So it might work.

plieven:
@sahlberg I actually do not know which penalty the FIONREAD ioctl has. I remember why where happy to successfully eliminate it from libnfs. As for the systems with lacking support of non-blocking sockets. Do we really support them? We at least try to set O_NONBLOCK on the socket unconditionally.

plieven:
@Chilledheart I generally agree that CPU power should be higher than network throughput, but our approaches so far processed all PDUs we received. This can invoke callbacks back into the application and we do not know how long they will take to complete. I would therefore propose that we consume everything in iscsi_read_from_socket until the socket would block or we reach a limit how much PDUs we want to batch at most. Than process all PDUs we received and then write everything that might have popped out in the outqueue.

plieven:
Let me know how you like this approach:

https://github.com/plieven/libiscsi/commit/963490a67cab3b209d28be5ceac3e4d86555e47d

I changed the approach to process the PDU directly after reading. Not 100% sure if this is better than first reading and then processing, but this way we can kill the inqueue (i do not know why it whas there at all maybe a leftover from the past) and I can break the while loop if the connection is terminated.

@sahlberg if there is a system which really is lacking non blocking socket support we could set MAX_BATCH_PDUS to 1 for this system. But please note that Linux sometimes even returns POLLIN if there is no data on the socket.

@Chilledheart can you benchmark this patch?

sahlberg:
It is a lot bigger change, so there is the danger of regressions, but I
like it.

I would like if you add a simple check that the socket is actually readable
at the start of the while loop and return if the socket is not readable.
Either using ioctl(FIONREAD)   or if you prefer, but it will be more code,
poll() without a timeout and check for POLLIN.

If we can keep it working even for cases where the platforms have broken or
no non-blocking support at all.
Those platforms are very niche  but if we can still support them at a small
cost, it would be nice.

Chilledheart:
@plieven it is cool to introduce batch submit. in your patch, the batch num is 32 which is sufficient to deal with most cases. 

But I don't think drop inqueue is a good idea, because fast scsi device such as nvme device can support high queue depth as 128 or so.

plieven:
@Chilledheart please have a look at the code before this patch. If I am not completely wrong, the inqueue has no effect. As soon as we read a full PDU, we add it to the inqueue and then directly pop it from the queue and process it. So why this inqueue?

plieven:
@sahlberg I would like to avoid this extra overhead. What do you think would it be feasible to track if the ioctl to set the socket to nonblocking succeeds. If yes we process up to 32 PDUs. If not we process a maximum of 1 which is identical to the old code.

sahlberg:
Some very niche implementations don't implement non-blocking I/O however if
you try to set the socket to non-blocking mode it is a NOOP and returns
success.
This makes porting of software easier as you dont have to remove the calls,
and error paths, for the call to make it non-blocking.
(most software probably work just fine even if they request but actually
does not get true non-blocking sockets)

These are not mainstream OSes though. I know of two; Amiga which I want to
support and some folks working on a research os which name escapes me.
Neither of these use cases are very important though.


Adding a poll() call should be near zero cost, at least on linux where the
context switch for these calls are very cheap and often <10us.
But I can live without this for now.    I may add it later via a configure
check/conditional if/when I get time to play with AROS (amiga) again.


I will try looking through the patch in more detail and merge it over the
weekend.
Can you give it some mileage with running on QEMU?   QEMU is the most
important consumer for libiscsi.

Chilledheart:
@sahlberg if poll()'s cost is less than 10us it looks fine to me. and if you want to test with qemu, I recommend malloc-based spdk iscsi target server which performs really great or lio target with nullblk backend which performs great as well.

@plieven sorry to late response but you know I am living in GMT+8 timezone so it is a bit hard to catch up with you. I don't have much free time in the work hours. Maybe you can create a new PR to refine this patch. sorry but I don't have much much time to improve with it.

Chilledheart:
@plieven  what's more, i think it is better to move batch submit detection logical to iscsi_which_events. in iscsi_which_events it is possible to deal inqueue and outqueue in both, maybe.

Chilledheart:
I've updated a new patch which uses poll to detect read events. I tested it a bit with iscsi-perf (and qemu tonight), and it performs really great. I witnessed a 20% performance boost, from 100K upto 120K. thank you, @sahlberg. It is likely that poll() scales better than read().

Chilledheart:
for reference, perf report for iscsi-perf is attached

screen shot 2016-12-17 at 2 42 18 pm

sahlberg:
That is very nice and compact.

In iscsi_tcp_poll_events get rid of the ret variable.
Also use { } for the if statements, even if the conditional block is a
single line.

plieven:
Yeah, that looks good.

I would limit to maximum number of iterations just for security reasons.

Peter

Chilledheart:
@sahlberg done

@plieven is this version better now?

sahlberg:
+ if (iscsi_tcp_poll_events(iscsi, POLLIN) <= 1) {

shouldn't this be

+ if (iscsi_tcp_poll_events(iscsi, POLLIN) <= 0) {

Chilledheart:
@sahlberg my bad. updated.

plieven:
maybe we should also mask pfd.events with the events we are interested in. otherwise poll might consume cpu cycles determinating if the socket is writeable while we are not at all interested in it at this place.

plieven:
this should be fixed as well: 

   for (int i = 0; i < READ_PDU_BATCH_COUNT; ++i) {
   ^
socket.c:947:3: note: use option -std=c99 or -std=gnu99 to compile your code

Chilledheart:
@plieven yes, we shouldn't waste cpu time on POLLOUT event here.

plieven:
we also need to break the loop if we logged out

plieven:
connected to iscsi://172.21.200.56/iqn.2001-05.com.equallogic:0-8a0906-1695c510a-3ad001f8b565232e-02convert/0
capacity is 8202240 blocks or 4199546880 byte (4005 MB)
performing SEQUENTIAL READ with 16 parallel requests
FIXED transfer size of 8 blocks (4096 byte)
infinite runtime - press CTRL-C to abort.

00:00:01 - lba 112168, iops current 14001 (54 MB/s), iops average 14001 (54 MB/s), in_flight 16, busy 0        
00:00:02 - lba 200784, iops current 11076 (43 MB/s), iops average 12539 (48 MB/s), in_flight 16, busy 0        
^C

finished.
libiscsi:2 logout successful [iqn.2001-05.com.equallogic:0-8a0906-1695c510a-3ad001f8b565232e-02convert]
libiscsi:1 iscsi_service_reconnect_if_loggedin. Can not reconnect right now.
 [iqn.2001-05.com.equallogic:0-8a0906-1695c510a-3ad001f8b565232e-02convert]
libiscsi:1 iscsi_service failed with : iscsi_service_reconnect_if_loggedin. Can not reconnect right now.
 [iqn.2001-05.com.equallogic:0-8a0906-1695c510a-3ad001f8b565232e-02convert]

plieven:
+                       if (!iscsi->is_loggedin) {
+                               break;
+                       }

Chilledheart:
i think this check is not necessary because in iscsi_tcp_which_events

```
static int
iscsi_tcp_which_events(struct iscsi_context *iscsi)
{
	int events = iscsi->is_connected ? POLLIN : POLLOUT;

	if (iscsi->pending_reconnect && iscsi->old_iscsi &&
		time(NULL) < iscsi->next_reconnect) {
		return 0;
	}
```

iscsi_tcp_which_events() won't return POLLIN is not connected

plieven:
i am talking about logged_in not connected. if i omit the check i run into the error above

plieven:
i also did two other mods

plieven:
+                       if (iscsi_tcp_poll_events(iscsi, POLLIN) != POLLIN) {
+                               break;
+                       }

if poll returns anything except POLLIN (e.g. POLLHUP, POLLERR etc.) we should also break

plieven:
if (errno == EINTR || errno == EAGAIN) {
				break;
			}

before iscsi_tcp_poll_events(iscsi, POLLIN.  if the recv was would have blocked we do not need to poll again

plieven:
my loop basically locks like this now:

+               int i;
+               for (i = 0; i < READ_PDU_BATCH_COUNT; ++i) {
+                       if (iscsi_read_from_socket(iscsi) != 0) {
+                               return iscsi_service_reconnect_if_loggedin(iscsi);
+                       }
+                       if (errno == EINTR || errno == EAGAIN || !iscsi->is_loggedin
+                           || iscsi_tcp_poll_events(iscsi, POLLIN) != POLLIN) {
+                               break;
+                       }
+                }

Chilledheart:
i see. there is more checks needed to deal with poll's POLLERR and POLLHUP. it is pretty bad to use two poll in single event polling routine.

plieven:
i am testing the above version with qemu and our internal backup software (which dumps snapshots from iscsi storages to FC arrays) now.

Chilledheart:
@plieven thank you, plieven. it is really kind of you to such check for this patch. but I think it requires more changes than I thought.

plieven:
i just came up with another idea to handle the situation. let me double think of it. maybe things are a lot easier.

plieven:
please have a look at #225

Chilledheart:
It is safe enough but not good at performance. FIONREAD get the number of immediate read at the right time ioctl() happens. There might be more later after read routine. What's more, useless syscalls are not great at performance at all.

Chilledheart:
Of course useless syscalls are referred in my patch.

plieven:
The idea was to not risk to starve in iscsi_read_from_socket for too long. The single FIONREAD ioctl at enter ensures that we read all PDUs of whom at least a single byte is already in the input buffer when we enter. This approach is at least a lot better than what we have at the moment. We have to find a tradeoff between latency and throughput.

plieven:
@sahlberg what do you think?

Chilledheart:
updated. fix a bug found in production environment.

lib/socket.c
```diff
@@ -653,14 +657,35 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
...
 
+// TODO(sahlberg) change this to 1 for AROS which doesn't support NONBLOCK io
+// actually
+#define READ_PDU_BATCH_COUNT 32
+
+static int iscsi_read_from_socket(struct iscsi_context *iscsi) {
+	int i;
+	for (i = 0; i < READ_PDU_BATCH_COUNT; ++i) {
+		struct iscsi_in_pdu *pdu;
+		int ret = iscsi_read_pdu_from_socket(iscsi, &pdu);
```
sahlberg: Why the PDU argument since it is not used ?

lib/socket.c
```diff
@@ -653,14 +657,35 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
...
+		/* NODATA */
+		if (ret == 0) {
+			break;
+		}
+		if (ret < 0) {
+			return -1;
+		}
```
sahlberg: Try a poll(POLLIN) check here. Or ioctl(FIONREAD).

That might be cheaper than calling iscsi_read_pdu_from_socket() again and relying on recv() failing.

sahlberg:
How does this patch perform compared to #225 ?
I have no strong feeling in either direction.  I like that #224 moves the
greedy loop out into its own function as the current read_from_socket is
already very complex.

plieven:
#224 tends to process more PDUs in an iscsi_service invocation than #225, but time spend per iscsi_service cal is greater. Performance seemms equal, at least for me. @Chilledheart can you test, you might have an iscsi target with a memory backend.

lib/socket.c
Chilledheart: I wanted to process pdu once read. maybe it should be removed.

lib/socket.c
Chilledheart: No, it is bad to do a poll here where you must deal with POLLERR and POLLHUP.

to use ioctl(FIONREAD), you must call it before every recv() which is not acceptable as well.

Chilledheart: I meant not acceptable for performance

Chilledheart:
if this patch can't process as many pdu as possible it means nothing since it doesn't fix the qemu performance issue.

we are doing tests (HA and perfomance tests, but not too strong) on a patched 1.9.0 libiscsi. It is with the same patch as this one. I'll push the source code later once the tests are passed.

sahlberg:
Ok.

Can you test plievens patch as well and compare ?

Chilledheart:
@sahlberg okay, let me do this comparison.

lib/socket.c
Chilledheart: I still don't think call ioctl(FIONREAD) is a good idea because currently 100G nic does 12.5G/s throughput which means 12.5 kb per us. you must miss many pdu during several syscalls after ioctl. for the case of 10G nic, you miss 1.25 kilobytes per us.

Chilledheart:
```host cpu: Intel(R) Xeon(R) CPU E5-2620 v2 @ 2.10GHz x 2
host mem: 64G
host kernel: 3.10.0-327.el7.centos.smartx.x86_64 (patched kernel from centos)

target: our target server on host
backend: memory
backend latency: 1us
backend queue depth: 32
(comment: tgt with memory backend is too slow with 56K iops, while our provides 130K iops, tested with iscsi-perf 1.18)

qemu: 2.8.0 with libiscsi 1.18 (static linked) on lo device
guest cpu: 1, passthrough
guest mem: 16G
guest kernel: 3.10.0-327.36.3.el7.centos.x86_64
guest configure: cache=none, aio=native, virtio-scsi

result:
             |  keyue+32 batch |  keyue+256 batch  |   plieven#225
-------------|--------------------------------------------------------
iscsi-perf   |  132.1K iops    |    133.3K iops    |   124.5K iops
4k+randread  |  77.9K iops     |    87.9K iops     |   86.3K iops
4k+randwrite |  35.6K iops     |    40.7K iops     |   38.1K iops

summary: 32 batch is too small for current server, 256 is better. plieven#225
patch is good enought currently, very close to 256 batch.
comment: 2.1Ghz cpu is still too slow to unleash full performance on libiscsi. previous tests on a i7 6700k 4.0G cpu, it shows up to 340K iops on a lo device
```
more details:

fio-perf-keyue
iscsi-perf-keyue

fio-perf-keyue-256
iscsi-perf-keyue-256


fio-perf-plieven
iscsi-perf-plieven

sitsofe:
I've tried out both patches on a Xeon E5-2640 v3 2.60GHz, using fio with a custom libiscsi backend going to a localhost SPDK iSCSI target with a memory based LUN. Both patches show improvement in IOPs (with the smartxworks patch being slightly higher) and a reduction in minimum and average case completion latencies when the target is fast enough. However, both patches seem to show higher worst case completion latencies (see the READ_Clat_pct15 column) compared to the unpatched libiscsi (with the smartxworks patch being the highest of all on average). Performance with patches while running swings about a lot making results erratic on shorter runs. Both patches can also show a maximum completion latency in the minutes range as if something became stuck - perhaps a bug in my libiscsi fio backend? The fio libiscsi backend alway shows lower IOPs than iscsi-perf no matter what settings I pass to fio so perhaps I need to work on the custom fio libiscsi backend some more.

Please see [libiscsi_batch_pdu.xlsx](https://github.com/sahlberg/libiscsi/files/675587/libiscsi_batch_pdu.xlsx) for a spreadsheet of results.

fio-plain: sahlberg/master
fio-sx: smartxworks/master
fio-pl: lieven/read_batch_pdu

fio job run was like
```
fio-$version --minimal --randseed=1 --thread=1 --ioengine=libiscsi --iscsi_url 'iscsi://127.0.0.1/iqn.2016-06.io.spdk:disk1/0' --filename 'iscsi\://127.0.0.1/iqn.2016-06.io.spdk\:disk1/0' --iodepth=128 --time_based --ramp_time=5 --rw=read --randseed=1 --name=$name-$b --runtime=120s --iodepth=128 --bs=$b
```

[fio results headings are described in the fio HOWTO 7.0 Terse Output](https://github.com/axboe/fio/blob/fio-2.16/HOWTO#L2174)

Chilledheart:
hi @sitsofe, it is really cool to test with fio libiscsi engine. It reveals many details and both patch are not ideal. (btw I didn't know the libiscsi fio before. I thought fio is GPL licensed, but I couldn't find the code online. so I wondered to write it myself recently)

but it is still different from the qemu's situation where causes high vmexit events if you returns too quick from iscsi_tcp_service() where the iscsi target is connected with virtio-scsi or virtio-blk via libiscsi. 

I think there is more than one problem left before us.

sitsofe:
@Chilledheart I only finished getting it going a few days ago (it's been a holiday project) and have yet to publish it but it's something I've been thinking about for a while due to what I do at work. It's essentially a port of @plieven's iscsi-perf to an fio ioengine but it's still rather rough around the edges as I was learning how to make things work.

Re high vmexits - I wonder if using `--iodepth_batch_complete_min=0 --iodepth_batch_complete_max=128` on fio would be more similar to what QEmu does?

Chilledheart:
@sitsofe sounds cool. you will find fio's linux aio ioengine helpful.

> Re high vmexits - I wonder if using --iodepth_batch_complete_min=0 --iodepth_batch_complete_max=128 on fio would be more similar to what QEmu does?
not so sure. you might want to test it you self. the high vmexits depends on cpu's frequency. for my case, I am running fio inside the vm and if iscsi targets provides more that 40K iops it degrades to 28K iops.  I saw a much greater number of vmexit compared to openiscsi with linux aio under the same environment. But when the fio running inside vm is  batch-submitted at 16 depth, the iops increases significantly.

Anyway I can't tell if your emulation is right.

sitsofe:
OK I identified a flaw in my fio libiscsi backend - completions aren't been returned in the order they happen which can lead to starvation. I'll post an update when its fixed.

sitsofe:
Here is an updated [libiscsi_batch_pdu_results.xlsx](https://github.com/sahlberg/libiscsi/files/679047/libiscsi_batch_pdu_results.xlsx) spreadsheet with completions being done in the order libiscsi's callbacks are fired. The name of a sheet indicates the queue depth and the maximum number of completions collected together after a `iscsi_tcp_service()` call.

* Both patches generally improve performance (higher average iops, lower average and minimum latencies) by a similar amount. The biggest gain seems to a 25% improvement in IOPs for `*-64-4k` in the `depth_64_max_64` sheet.
* On average the smartxworks/master patch provides higher IOPs for block sizes 512 and 64kbyte whereas the lieven/read_batch_pdu patch is better for 4kbyte blocks.
* Performance was better in all versions with a queue depth of 64 than with a queue depth of 128.
* Both patched versions struggled to outperform (or worse were slower) than the unpatched version when the queue depth was 64, I/Os were only completed up to 64 or only one at a time and the block size was 64kbytes.

sitsofe:
Another observation: both patches tended to reduce user CPU usage while increasing system CPU usage. The lieven/read_batch_pdu patch tended to have lower user CPU and higher system CPU than the smartxworks/master patch on the same workload.

sahlberg:
Thanks for running these tests Sistofe.

I think 4kb blocks is the most important metric as this is what we will
mostly see from QEMU.
While the differences are not dramatic, I think it is time to decide on
which of the two patches to merge as a starting point and to continue
working on.

I am leaning towards merging plievens patch at this point. I have no strong
feelings for either patch and think they are both fine. However plievens
patch is slightly better for 4kb blocks
and I want to merge one of them so that we can focus on building and
improving on it even further.


Comments?

Chilledheart:
Still, I don't favour the ioctl call's way, which is really unusual in network programming. And there is much fix to do in the tcp layer. I am afraid I will prefer to fork libiscsi and rewrite the tcp part rather than spend one month begging for merge it.

plieven:
I don't like the ioctl version also, but what I would like to avoid is to poll inside iscsi_service. Here we have a nested poll and we go round the error handling that the application might do with the results of its own poll before invoking iscsi_service. We could avoid poll and ioctl if we would hard code to avoid the loop for those OSs that allow to enable non-blocking I/O and then block on read if there is no data. Thats the only reason we added this.

Chilledheart:
I don't want to say more, but it is pretty common to deal different OS with different versions of function or event different routines.

It makes me silly to spend time arguing with how to write "perfect" or "ideal" code in os level. Our programs don't reply on it and just need practical ones. Thank you, guys. Have a nice time.

PR #225, opened 2016-12-20
Batch pdu read in function iscsi_tcp_service()

plieven:
iscsi_read_from_socket can currently only read one PDU in each iscsi_service invocation even
if there is more data available on the socket. This patch reads all PDUs available on
the socket when iscsi_read_from_socket is entered. It enqueues all complete read PDUs and
then processes them in order of arrival.

Suggested-by: Chilledheart <rwindz0@gmail.com>

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -567,99 +567,110 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 {
 	struct iscsi_in_pdu *in;
 	ssize_t data_size, count, padding_size;
+	int available;
 
-	if (iscsi->incoming == NULL) {
-		iscsi->incoming = iscsi_szmalloc(iscsi, sizeof(struct iscsi_in_pdu));
-		iscsi->incoming->hdr = iscsi_szmalloc(iscsi, ISCSI_RAW_HEADER_SIZE + ISCSI_DIGEST_SIZE);
-		if (iscsi->incoming == NULL) {
-			iscsi_set_error(iscsi, "Out-of-memory: failed to malloc iscsi_in_pdu");
-			return -1;
-		}
-	}
-	in = iscsi->incoming;
-
-	/* first we must read the header, including any digests */
-	if (in->hdr_pos < ISCSI_HEADER_SIZE) {
-		/* try to only read the header, the socket is nonblocking, so
-		 * no need to limit the read to what is available in the socket
-		 */
-		count = ISCSI_HEADER_SIZE - in->hdr_pos;
-		count = recv(iscsi->fd, &in->hdr[in->hdr_pos], count, 0);
-		if (count == 0) {
-			return -1;
-		}
-		if (count < 0) {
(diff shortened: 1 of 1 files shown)


lib/socket.c
```diff
@@ -567,100 +567,111 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
...
-	if (iscsi->incoming == NULL) {
-		iscsi->incoming = iscsi_szmalloc(iscsi, sizeof(struct iscsi_in_pdu));
-		iscsi->incoming->hdr = iscsi_szmalloc(iscsi, ISCSI_RAW_HEADER_SIZE + ISCSI_DIGEST_SIZE);
-		if (iscsi->incoming == NULL) {
-			iscsi_set_error(iscsi, "Out-of-memory: failed to malloc iscsi_in_pdu");
-			return -1;
-		}
+	/* check how much data is in the input buffer of the socket and read as many
+	 * PDUs as available */
+	if (ioctl(iscsi->fd, FIONREAD, &available) < 0) {
```
sahlberg: if (ioctl(iscsi->fd, FIONREAD, &available) < 0 || available == 0) {

lib/socket.c
```diff
@@ -567,100 +567,111 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
...
+			if (count < 0) {
+				if (errno == EINTR || errno == EAGAIN) {
+					break;
+				}
+				iscsi_set_error(iscsi, "read from socket failed, "
+					"errno:%d", errno);
+				return -1;
+			}
+			in->hdr_pos  += count;
+			available -= count;
```
sahlberg: Would need a check if available <= 0  here and abort.

Otherwise, if we have only reveiced exacly just the header but not the payload, we would still have available > 0 here
but as there is no more data to receive, lines 636 or 648 would return 0 and thus 650 would incorrectly think the socket has been clsoed.

lib/socket.c
plieven: Not in Linux. In Linux poll sometimes returns POLLIN even if the receive buffer is empty. In this case available would be 0, but read would not return 0 but -1 with errno = EWOULDBLOCK.

lib/socket.c
plieven: Same as above, if there are not bytes available in the socket read does not return 0 but -1 with EWOULDBLOCK. It happens quite often that available is < 0 at the end. read reads up to the full PDU. It might be that when the FIONREAD ioctl is send only part of the PDU is in the receive buffer, but the rest arrives while we are in the loop. Thus my commit message that we read all PDUs that are at least partly in the buffer. However, the FIONREAD approach guarantees that we avoid starvation in the loop and increased latency. As I explained earlier. We also might invoke a call back for each PDU and we have no idea how long that callback takes. So if we poll again are reissue FIONREAD ioctl we might stay in the loop quite a long time. I thus read first, put the PDUs in the inqueue (which finally has a meaining (again), and process all in a row.

PR #226, opened 2016-12-22
TESTS: exceed Maximum Descriptor List Length

ddiss:
ExtendedCopy.DescrLimits currently attempts to test Maximum Descriptor
List Length bounds checking by issuing a request with
(max_target_desc_count + 1 + max_segment_desc_count + 1) descriptors.

Maximum Descriptor List Length is a separately advertised field, that
may not be exceeded with the above logic, e.g. LIO advertises:
- Max Target Desc Count = 2
- Max Segment Desc Count = 1
- Max Desc List Len = 1024

Calculate the number of descriptors using the advertised Maximum
Descriptor List Length value, to explicitly test bounds checking.

### DIFF
--- a/test-tool/test_extendedcopy_descr_limits.c
+++ b/test-tool/test_extendedcopy_descr_limits.c
@@ -26,6 +26,7 @@
 #include "iscsi.h"
 #include "scsi-lowlevel.h"
 #include "iscsi-test-cu.h"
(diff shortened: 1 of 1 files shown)


ddiss:
Am seeing a valgrind error with this change - will fix and resubmit. Sorry for the noise.

PR #230, opened 2017-01-02
examples/dd: add XCOPY support

ddiss:
EXTENDED COPY can be triggered with the new --xcopy/-x parameter. When
invoked, (--max) EXTENDED COPY requests are dispatched in parallel, with
each request attempting to copy (--blocks) from source to destination.

Submitting for early feedback - I should probably split out the populate_X helpers into a library that can be shared with the test-tool code, as currently a lot of this is duplicate.

### DIFF
--- a/examples/iscsi-dd.c
+++ b/examples/iscsi-dd.c
@@ -39,19 +39,23 @@ struct client {
 	int src_lun;
 	int src_blocksize;
 	uint64_t src_num_blocks;
+	struct scsi_inquiry_device_designator src_tgt_desig;
 	uint64_t pos;
 
 	struct iscsi_context *dst_iscsi;
 	int dst_lun;
 	int dst_blocksize;
 	uint64_t dst_num_blocks;
+	struct scsi_inquiry_device_designator dst_tgt_desig;
 	int use_16_for_rw;
+	int use_xcopy;
 	int progress;
 	int ignore_errors;
 };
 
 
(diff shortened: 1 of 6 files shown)


sahlberg:
Very Nice.

Agree. I think the populate_ functions should go into the main library,
as well a new function iscsi_xcopy_async().

As this is new functionality, that does not block me merging the current
patch as it.

Your call, I can do either.

* merge as is now and you, me or someone else adds a proper
iscsi_xcopy_async() call afterwards
* another spin of the patch where you do these changes?

(
I don't think TGTD has XCOPY support yet, so I  might try adding it as this
will be a very nice tool to
test and verify with. This is very helpful for that.
)

ddiss:
Thanks for the feedback Ronnie.
I think some duplication of the populate_X functions will still be needed, as the tests are poking into obsolete / reserved fields. I'll whip up a quick iscsi_xcopy_async() change and resubmit with that, the common populate_X functions could then be done at a later date.

ddiss:
Changes since previous submission:
- add libiscsi RECEIVE COPY RESULTS and EXTENDED COPY async/sync helper functions
- validate blocks per I/O and max in flight parameters
  + check blocks per I/O value doesn't exceed XCOPY limits
- split out duplicate readcap code

sahlberg:
Awesome, thanks!

plieven:
dont we need to bump the api version?

PR #231, opened 2017-01-02
Batch pdu read in function iscsi_read_from_socket()

plieven:
iscsi_read_from_socket can currently only read one PDU in each iscsi_service invocation even
if there is more data available on the socket. This patch reads all PDUs until the socket
would block. It enqueues all complete read PDUs and then processes them in order of arrival.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -567,6 +567,9 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 {
 	struct iscsi_in_pdu *in;
 	ssize_t data_size, count, padding_size;
+	int waitpdu_len, inqueue_len = 0;
+
+	ISCSI_LIST_LENGTH(&iscsi->waitpdu, waitpdu_len);
 
 	do {
 		if (iscsi->incoming == NULL) {
@@ -638,7 +641,6 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 				}
 				count = recv(iscsi->fd, buf, count, 0);
 			}
-			
 			if (count == 0) {
 				/* remote side has closed the socket. */
 				return -1;
@@ -652,7 +654,6 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 						iscsi_get_error(iscsi));
 				return -1;
 			}
-
 			in->data_pos += count;
 		}
 
@@ -661,8 +662,9 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 		}
 
 		ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
+		inqueue_len++;
 		iscsi->incoming = NULL;
-	} while (iscsi->is_loggedin && iscsi->tcp_nonblocking);
+	} while (iscsi->tcp_nonblocking && inqueue_len < waitpdu_len && iscsi->is_loggedin);
 
 	while (iscsi->inqueue != NULL) {
 		struct iscsi_in_pdu *current = iscsi->inqueue;
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -567,9 +567,6 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 {
 	struct iscsi_in_pdu *in;
 	ssize_t data_size, count, padding_size;
-	int waitpdu_len, inqueue_len = 0;
-
-	ISCSI_LIST_LENGTH(&iscsi->waitpdu, waitpdu_len);
 
 	do {
 		if (iscsi->incoming == NULL) {
@@ -661,20 +658,13 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
 			break;
 		}
 
-		ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
-		inqueue_len++;
 		iscsi->incoming = NULL;
-	} while (iscsi->tcp_nonblocking && inqueue_len < waitpdu_len && iscsi->is_loggedin);
-
-	while (iscsi->inqueue != NULL) {
-		struct iscsi_in_pdu *current = iscsi->inqueue;
-
-		if (iscsi_process_pdu(iscsi, current) != 0) {
+		if (iscsi_process_pdu(iscsi, in) != 0) {
+			iscsi_free_iscsi_in_pdu(iscsi, in);
 			return -1;
 		}
-		ISCSI_LIST_REMOVE(&iscsi->inqueue, current);
-		iscsi_free_iscsi_in_pdu(iscsi, current);
-	}
+		iscsi_free_iscsi_in_pdu(iscsi, in);
+	} while (iscsi->tcp_nonblocking && iscsi->waitpdu && iscsi->is_loggedin);
 
 	return 0;
 }
@@ -988,16 +978,6 @@ iscsi_free_iscsi_in_pdu(struct iscsi_context *iscsi, struct iscsi_in_pdu *in)
 	in->data=NULL;
 	iscsi_sfree(iscsi, in);
 	in=NULL;
-}
-
-void
-iscsi_free_iscsi_inqueue(struct iscsi_context *iscsi, struct iscsi_in_pdu *inqueue)
-{
-	while (inqueue != NULL) {
-	      struct iscsi_in_pdu *next = inqueue->next;
-	      iscsi_free_iscsi_in_pdu(iscsi, inqueue);
-	      inqueue = next;
-	}
 }
 
 void iscsi_set_tcp_syncnt(struct iscsi_context *iscsi, int value)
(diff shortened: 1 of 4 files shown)


plieven:
@sitsofe can you test this one in your test suite as well please.

sitsofe:
@plieven  I'll give it a whirl when I next have access to the server I used. How does it differ from https://github.com/sahlberg/libiscsi/pull/225 ?

plieven:
@sitsofe it does not poll or use ioctl at all. It just reads until the socket would block. I record if setting the socket to non-blocking succeeds. If there are OSs around which succeed in setting to O_NONBLOCK, but then block at read this is a violation of the standard. But we might still set non blocking to false manually for those. When do you have next access to this machine?

sitsofe:
@plieven Most likely tomorrow and if not then definitely the day after.

plieven:
Great, thank you!

lib/socket.c
```diff
@@ -568,98 +568,101 @@ iscsi_read_from_socket(struct iscsi_context *iscsi)
...
 
-	ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
-	iscsi->incoming = NULL;
+		if (in->data_pos < data_size) {
+			break;
+		}
 
+		ISCSI_LIST_ADD_END(&iscsi->inqueue, in);
+		iscsi->incoming = NULL;
+	} while (iscsi->is_loggedin && iscsi->tcp_nonblocking);
```
sahlberg: At the end of this loop there is opportunity for a tiny optimization.

If the length of iscsi->waitpdu and iscsi->incoming are the same then except for any target initiated NOPs we should have a received any and all possible pdus from this socket and can abort early.

This avoids running the loop one more time just to fail with EAGAIN at the recs/readv. Just avoiding that recv/readv syscall will shave at least 10us off this function and thus the latency.

plieven:
Good idea, i already thought of checking if waitpdu->next is not NULL, but counting is even better.

The inqueue is always empty when we enter the function, right? So we can easily count it's length by the number of PDUs we added in the loop and the waitpdu queue can be counted before the loop.

As for target initiated messages there are also async messages, but I think there are also not that critical.

plieven:
great. when do you think you will have the results?

sitsofe:
Here are some preliminary results (unfortunately the build tested doesn't include e058e82): 
[libiscsi_batch_pdu_results_2_preliminary.xlsx](https://github.com/sahlberg/libiscsi/files/687831/libiscsi_batch_pdu_results_2_preliminary.xlsx) (waiting on depth_64_max_64). The new patch is pl2 and in general performance seems comparable to that of the previous patch. A quick glance show it is slightly worse than the previous pl patch at 512byte and 4kbyte blocks but faster at 64kbyte. Usually faster than unpatched libiscsi except on the depth_1_max_1 scenario...

plieven:
thanks a lot! the depth 1 stuff should be better with e058e82.

do you have a chance to test it as well?

Peter

sitsofe:
@plieven - I admire your enthusiasm :-)

[libiscsi_batch_pdu_results_3.xlsx](https://github.com/sahlberg/libiscsi/files/688621/libiscsi_batch_pdu_results_3.xlsx)

e058e82 shows strong (always top 3) results at high (128) queue depths. It struggles a bit at depth_64_max_1 and depth_1_max_1 but put in a very good showing for depth_1_max_1 blocksize 4k.
Playing with the pivot table source to look at average IOPs and then sorting on columns Test Name, Average - READ_IOPS makes it a bit easier to see which binary "won" a set of results. depth_64_max_64 only has results for two of the patches. Results still vary a lot on single tests though - not sure why.

fio-sx: smartxworks:master (roughtly https://github.com/sahlberg/libiscsi/pull/224/files )
fio-plain: libiscsi without batch patches.
fio-pl: FIONREAD version ((lieven/read_batch_pdu).
fio-pl2: b81e9a2 version (older lieven/read_batch_pdu2)
fio-pl3: e058e82 (lieven/read_batch_pdu2)

plieven:
Thank you, @sitsofe ;-)

Can you clearify what the 2 numbers in the naming scheme of the spread sheet mean?
depth_N_max_M ?

What is your impression which way should we go. fio-pl is out since the ioctl stuff seems to break Qemu.
I still wonder why fio-pl3 can be slower than fio-pl2. I see 2 possible reasons:

a) The counting of the waitqeue length can be expensive. (as adding an element at the end of a queue with only a head pointer) - there is room for optimization
b) the condition at the end of the do .. while loop is wrong. Can there be more PDUs than waitpdus. -> yes there can, especially for data in PDUs that are not final.

Let me add a change that I had in mind earlier. It illiminates a+b. However there is change for increased latency if the network throughput is higher than callback completion.

plieven:
@sitsofe can you benchmark eb7c1d9 as well?

sitsofe:
depth_N_max_M :

`N` is the maximum depth that fio will submit I/O to (`iodepth` - https://github.com/axboe/fio/blob/fio-2.16/HOWTO#L839 ). For example if `N` is 16 fio will submit 16 I/Os and then start waiting for completion. If only one I/O completes and is repeated then a single new I/O will be submit and fio will go back to waiting for completion. If 8 have completed and have been repeated then next time 8 new I/Os will be submitted before waiting for completion.

`M` is the *maximum* number of I/Os to complete/reap at any given time (`iodepth_batch_complete_max` which defaults to fio defaults to whatever `iodepth_batch_complete_min ` is which defaults to 1 - https://github.com/axboe/fio/blob/fio-2.16/HOWTO#L868). When fio starts waiting for a completion it won't stop waiting and thus won't submit any new I/O, only pump libiscsi to see if new completions have arrived. When the libiscsi callback is fired and notices its I/O has completed, it adds the I/O to a completion ring. The part of fio waiting for completion checks this ring after a pump to determine completions to be reaped but will only take up to `M` I/Os off it in one go. So if `N` is 16 but `M` is 1, 16 I/Os are queued and sent but if they all arrive at the same time only one is taken per completion function call potentially leading to only one new I/O being submitted at a time because the other 15 entries hadn't been repeated. Each time fio waits for completion it will pump libiscsi at least once.

It doesn't make sense for `M` to be bigger than `N` (I think fio may internally adjust `M` if it finds such a scenario is requested).

sitsofe:
@plieven I'm tempted by fio-pl3 as it tends to do well at 4k blocksizes. I guess the snags in what I'm doing are:

* Currently I'm busy waiting completions - polling may show a different result.
* I'm only testing one scenario (fio-libiscsi -> target) whereas I guess many will care far more about the QEmu scenario (internal io -> qemu-libiscsi -> target).
* I've been focussing on sequential reads and haven't done random reads, writes or mixtures.
* It's possible to tell fio to wait until it has at least X completions by using `iodepth_batch_complete_min`. Setting this to 4 or above and using `gtod_reduce=1` (which produces far less statistics) is the only time I've seen fio-libiscsi produce a higher IOPs than iscsi-perf. Doing this may change results again...
* The results I get seem to fluctuate about for some reason. I can't tell if this is due to the environment, my patches to fio, libiscsi or the target.

The problem is runs take a while to finish as it stands - running each test three times for two minutes means the time adds up unfortunately so I'm loathe to test any more combinations for now (new commits should be OK until the start of next week).

I'll try commit eb7c1d9 now...

plieven:
@sitsofe thank you very much for your efforts. I personally would like to close this as soon as possible and focus on other things. If we have evidence which of pl2,3,4 is best, I would ask Ronnie to merge it. I don't wont to optimize it too much.

sitsofe:
Latest results: 
[libiscsi_batch_pdu_results_4.xlsx](https://github.com/sahlberg/libiscsi/files/690343/libiscsi_batch_pdu_results_4.xlsx)

Latest patch (pl4) still does very well at high queue depths and sometimes does a little better than previous patches in its lineage at lower queue depths. Still struggles on depth_64_max_64 and depth_1_max_1 on non-4kbyte block sizes but seems a little bit better on average (or if you take the best result of the three runs).

sahlberg:
Sistofe, Many many thanks for all your effort in doing these tests.
Peter, thanks for your work on these patches as well.


I think, as QEMU is the most important consumer of this library, that
performance for 4kb block sizes is what is most important and it trumps
performance for non-4kb.

Since there are no dramatic changes in the last 3 iterations of plievens
patch, we might be at or very close to how good we can get 4kb performance
with the current approach, i.e. tweaking how many PDUs we process for each
POLLIN event, and that we are now chasing diminishing returns.


Do you agree that is time to merge this latest patch and declare the
victory ? I think so.

plieven:
yes, please go ahead. please also have a look at the other PR i have sent.

Peter

sahlberg:
Merged, thanks!

PR #233, opened 2017-01-05
fix crc32c header checksums and some bugs

plieven:
@sahlberg as discussed
@shterman as far as i read the iser code it does not support header checksums. we should abort then if the target requests HeaderDigest=CRC32C


sahlberg:
Merged. Thanks.

plieven:
commit 1a552a8 got lost.

sahlberg:
fixed

Issue #178, opened 2015-07-19
Overflow in iscsi-perf.c


mykaul:
In fill_read_queue(), I believe that num_blocks is an int, and the calculation:
num_blocks = client->num_blocks - client->pos;

Might overflow a bit (as client->num_blocks is a 64bit unsigned number and client->pos may be zero in sequential IO), so num_blocks may end up as a negative value.
Passed on to other functions, it may be well over the limit we wanted. Certainly if it's a negative number, the check:
if (num_blocks > blocks_per_io) {
                         num_blocks = blocks_per_io;
                 }

Will not work.

mykaul:
In my case, I've tried against a ~2TB drive:
capacity is 4092205056 blocks or 2095208988672 byte (1998147 MB)

mykaul:
Proposed fix (not sure if it's good enough):
yanivk@vd-yanivk7:~/libiscsi (master)> git diff ./utils/iscsi-perf.c
diff --git a/utils/iscsi-perf.c b/utils/iscsi-perf.c
index a27d50a..a40f905 100644
--- a/utils/iscsi-perf.c
+++ b/utils/iscsi-perf.c
@@ -40,7 +40,7 @@
 const char *initiator = "iqn.2010-11.libiscsi:iscsi-perf";
 int proc_alarm = 0;
 int max_in_flight = 32;
-int blocks_per_io = 8;
+uint blocks_per_io = 8;
 uint64_t runtime = 0;
 uint64_t finished = 0;
 
@@ -178,7 +178,7 @@ out:
 
 void fill_read_queue(struct client *client)
 {
-       int num_blocks;
+       uint num_blocks;
 
        if (finished) return;
 


I'm not sure if it should not be a uint64_t too.

PR #234, opened 2017-01-09
iscsi-perf: fix overflow issue in fill_read_queue

optimistyzy:
If with int defintion, this statement:

num_blocks = client->num_blocks - client->pos;

May not get the exact value we want.

Signed-off-by: optimistyzy <optimistyzy@gmail.com>

### DIFF
--- a/utils/iscsi-perf.c
+++ b/utils/iscsi-perf.c
@@ -186,7 +186,7 @@ out:
 
 void fill_read_queue(struct client *client)
 {
-	int num_blocks;
+	int64_t num_blocks;
 
 	if (finished) return;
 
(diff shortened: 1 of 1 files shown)


sitsofe:
Mentioning @plieven on this too.

@optimistyzy's patch seems required because if `client->num_blocks` is large (imagine a 2tbyte target with 512 byte blocks) and we happen to pick the last block because we asked for a random sequence, `client->pos` can still be 0 (because we it was only recently initialized via memset) causing `client->num_blocks - 0` (2147483648) to be interpreted as negative when stored in num_blocks.

optimistyzy:
Also it is previously reported an issue in https://github.com/sahlberg/libiscsi/issues/178

sahlberg:
Merged. thanks.

PR #236, opened 2017-01-23
sync: fix return value for various sync commands

plieven:
all sync commands that return an integer value are
supposed to return a negative value on error.
However, state.status is positive non-zero on error.
Fix this by returning -1 if the state.status is
not SCSI_STATUS_GOOD.

Reported-by: Sitsofe Wheeler <sitsofe@yahoo.com>

### DIFF
--- a/lib/sync.c
+++ b/lib/sync.c
@@ -111,7 +111,7 @@ iscsi_connect_sync(struct iscsi_context *iscsi, const char *portal)
 	/* clear connect_data so it doesnt point to our stack */
 	iscsi->connect_data = NULL;
 
-	return state.status;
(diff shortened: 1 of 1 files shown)


plieven:
checked Qemu and libiscsi utils/examples. They all check for != 0. So no problem here.

sitsofe:
That's the approach I'm going to take as well. Do you think it's worth bumping up the API version for this too?

plieven:
I think its not necessary. We actually just fixed the routines to return the correct value as in the docs. If you want to be compatbile check for != 0.

sahlberg:
merged. thanks

Issue #243, opened 2017-03-03
Visual Studio project file


sahlberg:
I need to add a VS project file, just like libnfs, so that we can build and test using vs online.

Otherwise the windows support will continue to bitrot :-(

sitsofe:
Something that has started to help the fio project is the [warning advert] [appveyor](https://www.appveyor.com/) CI service. fio was using [travis](https://travis-ci.org/axboe/fio/builds/204854099) already and is relatively self contained so it wasn't onerous to add an [appveyor build](https://github.com/axboe/fio/blob/master/appveyor.yml). Here's the output of such a run: https://ci.appveyor.com/project/axboe/fio/build/1.0.9/job/xll06594rjpqa147 .

sahlberg:
I have added an initial visual studio project definition in bfd5bd944a206e4eea53df0d0a36aed0e4417237
I also fixed all the build errors, of which there were many that had accululated.

There are still a bunch of warnings but I think they are mostly benign. I will address them in a follow up patch though.


Can you try if this build works for you?

sahlberg:
I will take a look at appveyor. Thanks for the suggestion!

sitsofe:
Hi Ronnie,

I've given it a whirl. VS 2015 Community edition complains about the redefinition of `sprintf`. Commenting that out allowed the solution to build but I when I tried to link against the library using mingw I got back errors like:

```
[...]

PR #253, opened 2017-11-29
Fix compilation with VS2017

crawfxrd:
The primary issue is that in MSVC 14.00 (VS2015) Microsoft added `snprintf` as a function to the standard library and prevents users from defining it to something else (typically this was `_snprintf`). So, only define it when using `_MSC_VER` < 1900.

Other changes are:
- Fix macro definition of `dup2`
- Add macro for `getpid`
- Add function definition for `win32_dup`
- Add missing `EXTERN`s
- Replace `WIN32` with `_WIN32`


sitsofe:
@crawfxrd Since you're here could you cast your eyes over #243 ? There have been a few unresolved issues getting libiscsi going on Windows and if you're able to help...

sahlberg:
Merged, thanks.

crawfxrd:
@sitsofe Not sure I can be of help. I am working to rebase our internal fork on 1.17.0 (previously based on 0e0073e, but history wasn't preserved). We haven't had any issues with it, but we do have local modifications.

PR #254, opened 2017-12-04
PDU cancellation fixes for sync connect calls

franciozzy:
These address a segfault which may happen if the event_poll() return early (eg. timeout) for sync connect calls. On those cases, it is possible for a context to have PDUs queued which point to a local stack which have returned.


JonKohler:
Good job, Felipe!

PR #273, opened 2018-09-30
Isid fix

franciozzy:
Fixes a serious bug around ISID generation which results in reconnect attempts using the same ISID.


franciozzy:
@sitsofe Great question... I have no idea what I've done that caused it. Just fixed the PR. Thanks for pointing it out.

sahlberg:
merged, thanks

PR #274, opened 2018-10-04
Fix warnings from Coverity and GCC 8

bonzini:
This fixes various memory leaks reported by Coverity, as well as new warnings in GCC 8 (sometimes by disabling them...).


sahlberg:
Merged. Thanks!

PR #275, opened 2018-10-05
Call srand() only once

franciozzy:
I would like to get some comments on this approach. I can't really test WIN32, so not sure what's the best practice here. Also, I didn't run into any bugs here, just fixing the code based on not being entirely happy with the last fixes (that I wrote...).

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -28,6 +28,8 @@
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


franciozzy:
Gentle ping. :)

sahlberg:
merged, thanks!

PR #286, opened 2019-02-21
Replace file variables with .dir-locals.el

crawfxrd:
For Emacs users, use the same settings for all files in `c-mode`.


sahlberg:
Merged. thanks!

Issue #289, opened 2019-04-15
test-tool regression: 16435a917d671c3fca4ef4a3372eaf8db4492e2b breaks MPIO


ddiss:
Program received signal SIGSEGV, Segmentation fault.
iscsi_service (iscsi=0x0, revents=5) at socket.c:999
999             return iscsi->drv->service(iscsi, revents);
(gdb) bt
#0  iscsi_service (iscsi=0x0, revents=5) at socket.c:999
#1  0x0000000000405142 in send_scsi_command (sdev=sdev@entry=0x675010, task=0x6779b0,
    d=d@entry=0x0) at iscsi-support.c:323
#2  0x0000000000405f4a in prin_read_keys (sdev=sdev@entry=0x675010, tp=tp@entry=0x7fffffffdd90,
    rkp=rkp@entry=0x7fffffffdd98, allocation_len=allocation_len@entry=16384)
    at iscsi-support.c:683
#3  0x0000000000404882 in clear_pr (sdev=0x675010) at iscsi-test-cu.c:780
#4  0x0000000000403bff in main (argc=<optimized out>, argv=<optimized out>)
    at iscsi-test-cu.c:1248

This appears to be due to the NOP responder check in send_scsi_command(), which is consuming the mp_sds[] array before connect_scsi_device() has been run on each path :

1241         for (i = 0; i < mp_num_sds; i++) {
1242                 res = connect_scsi_device(mp_sds[i], initiatorname1);
1243                 if (res < 0) {
1244                         fprintf(stderr,
1245                                 "Failed to connect to SCSI device %d\n", i);
1246                         goto err_sds_free;
1247                 }
1248                 if (clear_pr(mp_sds[i]) < 0) {
1249                         printf("One or more persistent reservations keys have been registered\n");              
1250                         return -1;
1251                 }
1252         }

clear_pr() calls through to send_scsi_command() which does a:

 315                         /* Device [0] is where we are doing all the I/O
 316                          * so it will always work the socket and respond
 317                          * to NOPs from the target.
 318                          * But we need to trigger a service event every
 319                          * now and then on all the other devices to ensure that
 320                          * we detect and respond to any NOPs.
 321                          */
 322                         for (i = 1; i < mp_num_sds; i++) {
 323                                 iscsi_service(mp_sds[i]->iscsi_ctx, POLLIN|POLLOUT);
 324                         }

Fix to follow...

ddiss:
[...]

PR #290, opened 2019-04-15
test-tool: fix MPIO clear_pr() behaviour

ddiss:
clear_pr() can be called before connect_scsi_device() has been invoked
on each MPIO sdev. This results in a NULL pointer dereference in the
send_scsi_command() MPIO NOP-responder code-path.

Fix this by invoking clear_pr() after the MPIO connect_scsi_device()
loop has run. clear_pr() also only needs to be invoked via a single
path, so do so via the first path only.

Link: https://github.com/sahlberg/libiscsi/issues/289
Fixes: 16435a9 ("iscsi-test-cu: Improve persistent reservation clearing")

### DIFF
--- a/test-tool/iscsi-test-cu.c
+++ b/test-tool/iscsi-test-cu.c
@@ -1245,11 +1245,12 @@ main(int argc, char *argv[])
                                 "Failed to connect to SCSI device %d\n", i);
                         goto err_sds_free;
                 }
-                if (clear_pr(mp_sds[i]) < 0) {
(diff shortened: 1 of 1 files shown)


ddiss:
@bvanassche : would be great if you could also take a quick look here.

bvanassche:
This patch looks good to me.

ddiss:
@sahlberg ping - would be good to get the MPIO tests working again.

sahlberg:
Merged, thanks.

Sorry for forgetting this one.

PR #291, opened 2019-06-12
Added support for iSER related iSCSI keys

sumitrai:
Added support for negotiating below keys:
RDMAExtensions, TargetRecvDataSegmentLength, and
InitiatorRecvDataSegmentLength.

These are required to support iSER. See RFC5046 Section 6.


sahlberg:
Merged, thanks.

PR #300, opened 2019-09-17
test-tool: add and link against libiscsipriv convenience library

ddiss:
As a follow up to https://github.com/sahlberg/libiscsi/pull/299 , this patchset sees iscsi-test-cu statically link against a new libiscsipriv library, which allows tests to now make full use of internal libiscsi functionality for PDU mainipulation etc.

Only a single test has been changed to use the newly available symbols, but further tests for SendTargets and Nop-Out will follow.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -1,18 +1,37 @@
-lib_LTLIBRARIES = libiscsi.la
+# test-tool needs to access private symbols, so build a noinst convenience
+# library.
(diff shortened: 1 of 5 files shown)


ddiss:
FWIW, I'm far from an autotools expert, so please check closely for silly Makefile mistakes.

sahlberg:
not an expert on autotools either  but it looks sane

PR #301, opened 2019-09-18
test-tool: add iSCSI SendTargets, Nop-Out and CHAP auth tests

ddiss:


ddiss:
I think it makes sense to merge the two SendTargets tests into a single source file, will do that and push the result here.

ddiss:
> I think it makes sense to merge the two SendTargets tests into a single source file, will do that and push the result here.

Done.

ddiss:
I've tacked on a couple of extra CHAP_A tests.

PR #306, opened 2019-11-13
Improve robustness of the code that restores .queue_pdu()

bvanassche:
In my tests I hit a stack overflow caused by infinite recursion of
.queue_pdu() calls, caused by .queue_pdu() not being restored in an
error path. Make the approach for restoring .queue_pdu() more robust.

Signed-off-by: Bart Van Assche <bvanassche@acm.org>


bvanassche:
Ping?

PR #307, opened 2019-12-09
four little patches

qsqx:


include/iser-private.h
```diff
@@ -158,7 +158,7 @@ struct iser_tx_desc {
 struct iser_cm_hdr {
 	uint8_t      flags;
 	uint8_t      rsvd[3];
-} __packed;
+} __attribute__((packed));
```
bvanassche: Applying __attribute__((packed)) to an entire data structure can make gcc 9 unhappy. Since __attribute__((packed)) is not necessary for this data structure, please consider removing "__packed" entirely. See also https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/Warning-Options.html#index-Waddress-of-packed-member

qsqx: Ok, it's unnecessary and I've removed it.

PR #309, opened 2020-01-28
Fix data segment length comparison to unsigned long

yoyoyopcp:
In logic.c, data segment parameters in the text segment are converted to
signed longs.  Changing from strtol -> strtoul fixes compiler errors on
certain platforms that warn against comparing a signed long with
uint32_t using MIN.


sahlberg:
Merged. Thanks.

PR #311, opened 2020-02-14
Simplify the logic that determines when to send headers

iammattcoleman:
The prior condition could be summarized as:
```
if ((first && second) || second) {
```

This will always evaluate to `second`:

```
((true && true) || true) == true
((false && true) || true) == true
((true && false) || false) == false
((false && false) || false) == false
```


bvanassche:
Thanks for the patch!

PR #314, opened 2020-02-21
examples/iscsi-dd: add usage and performance output

ddiss:
iscsi-dd is quite useful for comparing XCOPY offloaded and regular READ/WRITE copy performance. This patchset makes iscsi-dd a little more user friendly for this use-case.


bvanassche:
Thanks for the patches!

PR #315, opened 2020-02-24
examples/iscsi-dd: check for clock_gettime() errors

ddiss:

### DIFF
--- a/examples/iscsi-dd.c
+++ b/examples/iscsi-dd.c
@@ -571,7 +571,7 @@ int main(int argc, char *argv[])
 	struct client client;
 	struct timespec start_time;
 	struct timespec end_time;
-
+	int gettime_ret;
 	static struct option long_options[] = {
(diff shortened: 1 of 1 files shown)


examples/iscsi-dd.c
```diff
@@ -758,7 +762,11 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	clock_gettime(CLOCK_MONOTONIC, &end_time);
+	ret = clock_gettime(CLOCK_MONOTONIC, &end_time);
+	if (ret < 0) {
+		fprintf(stderr, "clock_gettime(CLOCK_MONOTONIC) failed\n");
+		exit(10);
+	}
```
bvanassche: Does the action taken if clock_gettime() fails have to be this drastic? How about modifying show_perf() such that it does not show the transfer speed if clock_gettime() fails?

ddiss: Good idea - fixed.

PR #318, opened 2020-03-04
specify number of blocks in write_same command

neutronsharc:
If "number of blocks" is 0, write_same will not erase the LBA range and
read-verify will fail.

### DIFF
--- a/test-tool/test_writesame10_unmap_until_end.c
+++ b/test-tool/test_writesame10_unmap_until_end.c
@@ -53,9 +53,11 @@ test_writesame10_unmap_until_end(void)
                         i * block_size, block_size, 0, 0, 0, 0, 0, scratch,
                         EXPECT_STATUS_GOOD);
 
+                // write 0-buffer to entire range.
+                memset(scratch, 0, block_size * i);
                 logging(LOG_VERBOSE, "Unmap %d blocks using WRITESAME10", i);
                 WRITESAME10(sd, num_blocks - i,
-                            block_size, 0, 0, 1, 0, 0, scratch,
+                            block_size, i, 0, 1, 0, 0, scratch,
                             EXPECT_STATUS_GOOD);
 
                 if (rc16->lbprz) {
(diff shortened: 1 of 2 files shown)


bvanassche:
Thanks for the patch!

sitsofe:
@neutronsharc I'm unsure about this change... If you read https://www.t10.org/pipermail/t10/2016-May/017985.html it says that if the device supports zero in the NUMBER OF LOGICAL BLOCKS field for WRITE SAME then the behaviour is specified and there's a check that the device supports this behaviour done earlier in the code (`if (inq_bl->wsnz)`). Do you know if there's a part of a newer spec that contradicts this?

Mentioning @bvanassche for additional comment.

sitsofe:
Digging through an SBC spec I see it says this:

> If the NUMBER OF LOGICAL BLOCKS field is set to zero and:
> a) the WSNZ bit is set to zero (see 6.6.4), then the number of contiguous logical blocks that are requested to be unmapped or written includes all of the logical blocks starting with the LBA specified in the LOGICAL BLOCK ADDRESS field and ending with the last logical block on the medium; or

Later on it says this:

> A WSNZ bit set to zero indicates that the device server may or may not support a value of zero in the NUMBER OF LOGICAL BLOCKS field of the WRITE SAME commands.

If a server is in the does not support a value of zero case shouldn't it throw an error rather than silently doing nothing?

bvanassche:
Thanks @sitsofe for having reported this. I have reverted commit dfff7e9d16bb.

sitsofe:
@bvanassche I've been wondering about the wording (which seems extremely careful): "A WSNZ bit set to zero indicates that the device server may or may not support a value of zero in the NUMBER OF LOGICAL BLOCKS field of the WRITE SAME commands." . What do you think the spec writers meant by "may or may not support a value of zero"? They were clearly trying to be accepting of certain implementations but which ones? Is there a chance they want to accept implementations that accept 0 but then do nothing? If so then it would render WRITE SAME with zero blocks totally useless...

bvanassche:
In SBC-3 r36 I found the following text: _If the WSNZ bit is set to zero, then a NUMBER OF LOGICAL BLOCKS field set to zero specifies that the number of contiguous logical blocks that are requested to be unmapped or written includes all of the logical blocks starting with the LBA specified in the LOGICAL BLOCK ADDRESS field to the last logical block on the medium._

This is what I found in SBC-4 r22: _If the NUMBER OF LOGICAL BLOCKS field is set to zero and the WSNZ bit is set to zero (see 6.6.4), then the number of contiguous logical blocks that are requested to be unmapped or written includes all of the logical blocks starting with the LBA specified in the LOGICAL BLOCK ADDRESS field and ending with the last logical block on the medium._

PR #319, opened 2020-03-11
Iser reconnect bugfix

pacepi:
This PR includes 2 patches about iSER reconnection:

1. iser: fix crash for sending pdu during reconnecting
2. iser: fix hang in rdma_destroy_id

### DIFF
--- a/lib/iser.c
+++ b/lib/iser.c
@@ -711,6 +711,13 @@ iscsi_iser_send_pdu(struct iscsi_context *iscsi, struct iscsi_pdu *pdu) {
 	iscsi_pdu_set_expstatsn(pdu, iscsi->statsn + 1);
 	ISCSI_LIST_ADD_END(&iscsi->waitpdu, pdu);
 
+	/* because of async reconnection, before reconnecting successfully,
+	   the argument 'iscsi' is the 'old_iscsi' with a empty connection.
+	 */
+	if (!iser_conn) {
+		return 0;
+	}
+
 	if (iser_initialize_headers(iser_pdu, iser_conn)) {
 		iscsi_set_error(iscsi, "initialize headers Failed\n");
 		return -1;
--- a/lib/iser.c
+++ b/lib/iser.c
@@ -201,7 +201,6 @@ iser_free_iser_conn_res(struct iser_conn *iser_conn, bool destroy)
 	if (destroy) {
 
 		if (iser_conn->cmthread) {
-			pthread_cancel(iser_conn->cmthread);
 			pthread_join(iser_conn->cmthread, NULL);
 			iser_conn->cmthread = 0;
 		}


pacepi:
@sahlberg @bvanassche 
Fix 2 issue for iSER:
1, fix crash for sending pdu during reconnecting 
2, fix hang in rdma_destroy_id

Could you please review this PR?

include/iser-private.h
```diff
@@ -195,6 +195,7 @@ struct iser_conn {
 	int                          qp_max_recv_dtos;
 	int                          min_posted_rx;
 	uint16_t                     max_cmds;
+	uint8_t                      cmthread_should_run;
```
bvanassche: Please use a mutex (pthread_mutex_t) + condition variable (pthread_cond_t) in addition to this flag.

lib/iser.c
```diff
@@ -55,6 +55,8 @@
 
 /* the  name  can  be up to 16 bytes long, including the terminating null byte*/
 #define ISER_CM_THREAD_NAME "iscsi_cm_thread"
+/* let the cmthread shutdown gracefully, wait at most 1000ms */
```
bvanassche: Why is this timeout necessary? What would happen if the code below is modified such that it waits indefinitely?

lib/iser.c
```diff
@@ -201,7 +203,17 @@ iser_free_iser_conn_res(struct iser_conn *iser_conn, bool destroy)
...
+			int retry = ISER_CM_THREAD_WAIT;
+
+			iser_conn->cmthread_should_run = 0;
+			while ((retry-- > 0) && (iser_conn->conn_state == CONN_ESTABLISHED)) {
+				usleep(1);
+			}
+
+			if (iser_conn->conn_state == CONN_ESTABLISHED) {
+				ISCSI_LOG(iscsi, 2, "iser_free_iser_conn_res: try to kill cm thread after destroy qp %dms", ISER_CM_THREAD_WAIT);
+				pthread_cancel(iser_conn->cmthread);
```
bvanassche: The commit message explains why calling pthread_cancel() can cause trouble. So please remove all pthread_cancel() calls.

lib/iser.c
```diff
@@ -711,6 +711,14 @@ iscsi_iser_send_pdu(struct iscsi_context *iscsi, struct iscsi_pdu *pdu) {
 	iscsi_pdu_set_expstatsn(pdu, iscsi->statsn + 1);
 	ISCSI_LIST_ADD_END(&iscsi->waitpdu, pdu);
 
+	/* because of async reconnection, before reconnecting successfully,
+	   the argument 'iscsi' is the 'old_iscsi' with a empty connection.
+	 */
+	if (!iser_conn) {
+		ISCSI_LOG(iscsi, 2, "iscsi_iser_send_pdu: queuing pdu to old_iscsi while reconnecting");
```
bvanassche: Please change this message into something that is comprehensible for users who are not familiar with the libiscsi source code.

include/iser-private.h
bvanassche: On second thought, I'm not sure introducing a pthread mutex and condition variable is essential.

pacepi: Because libiscsi uses RDMA RC, client/server need send/receive DisconnectRequest & DisconnectReply. During disconnecting, the network may hit packet dropping case. 
I'm not sure the cm thread can receive the disconnect event or not, it's hard to emulate this case.
On the other hand, qemu locks mutex, then reconnect, finally unlocks mutex. If we blocks in reconnect for a long time, qemu will hit lock racing. So I add a timeout here.

In fact, I tested for 100 times, 1 case actually hit a timeout.

lib/iser.c
pacepi: There is already a log msg in iscsi_scsi_command_async, how about removing log here?

pacepi:
@bvanassche Rework this PR, could you please review this PR?

bvanassche:
A slightly modified version of these patches has been applied. Thanks!

PR #324, opened 2020-04-06
several fix relevant to iser connection

qsqx:


lib/iser.c
```diff
@@ -108,16 +103,27 @@ iscsi_iser_service(struct iscsi_context *iscsi, int revents)
 	int ret = 0;
 	struct iser_conn *iser_conn = iscsi->opaque;
 
+	if (iscsi->pending_reconnect) {
+		if (time(NULL) >= iscsi->next_reconnect) {
```
bvanassche, line 108: Please use clock_gettime(CLOCK_MONOTONIC, ...) instead of a function like time() that returns wall clock time

qsqx: These lines are consistent with those in iscsi_tcp_service() (socket.c:903), and next_reconnect is calulated by ```iscsi->next_reconnect = time(NULL) + backoff;``` in iscsi_reconnect_cb() (connect.c:300).
It might be better to modify them together in a separate commit if it matters.

qsqx:
> No period (.) at the end of a patch subject please.

fixed.

bvanassche:
Thanks for the patches!

lib/iser.c
bvanassche: That sounds fine to me.

PR #328, opened 2020-05-09
socket: fix rewrite cmdsn of immediate pdus

qsqx:
Cmdsn of a data-out pdu struct is less than `expcmdsn` since it's from its
cmd pdu. A data-out pdu doesn't carray a cmdsn on the wire actually, so it
doesn't matter to itself, but if we rewrite the cmdsn of a immediate pdu with
it, it will cause an error.

Related error logs:
libiscsi: iscsi_write_to_socket: outqueue[0]->cmdsn < expcmdsn (3648bab5 < 3648bab9) opcode 00 [iqn.2003-01.org.linux-iscsi.tgt0]
libiscsi: reconnect initiated [iqn.2003-01.org.linux-iscsi.tgt0]
libiscsi: connecting to portal 127.0.0.1 [iqn.2003-01.org.linux-iscsi.tgt0]
libiscsi: connection established (127.0.0.1:62404 -> 127.0.0.1) [iqn.2003-01.org.linux-iscsi.tgt0]

Signed-off-by: wanghonghao <wanghonghao@bytedance.com>

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -105,11 +105,19 @@ iscsi_add_to_outqueue(struct iscsi_context *iscsi, struct iscsi_pdu *pdu)
 	/* queue pdus in ascending order of CmdSN. 
 	 * ensure that pakets with the same CmdSN are kept in FIFO order.
 	 * immediate PDUs are queued in front of queue with the CmdSN
-	 * of the first element in the outqueue.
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #331, opened 2020-06-17
test-tool: Use block_size instead of hardcoded 512 bytes

heroin-moose:
This commit fixes some tests (like ProutReserve) on pure 4k devices.


bvanassche:
Thanks for the patch!

PR #332, opened 2020-07-01
Fix cancellation handling

YongjiXie:
This patchset fixes some bugs on cancellation handling. The first patch tries to plug the cmdsn gap when the pdus are cancelled before sending out. The second one remove a check to enable cancellation during login. The last two patches make the cancellation handling work during reconnecting.


bvanassche:
Thanks for the patches!

PR #333, opened 2020-08-07
Iscsi-dd cleanup

ddiss:
remove some duplicate code

### DIFF
--- a/examples/iscsi-dd.c
+++ b/examples/iscsi-dd.c
@@ -32,22 +32,23 @@ const char *initiator = "iqn.2010-11.ronnie:iscsi-inq";
 uint32_t max_in_flight = 50;
 uint32_t blocks_per_io = 200;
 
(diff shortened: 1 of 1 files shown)


examples/iscsi-dd.c
```diff
@@ -32,22 +32,23 @@ const char *initiator = "iqn.2010-11.ronnie:iscsi-inq";
 uint32_t max_in_flight = 50;
 uint32_t blocks_per_io = 200;
 
+struct src_dst {
```
bvanassche, line 35: How about renaming 'struct src_dst' into 'struct iscsi_endpoint'? Otherwise this patch looks good to me.

ddiss: That's a much nicer name, thanks.

bvanassche:
Thanks for the patches!

PR #334, opened 2020-08-10
test-tool: Use block_size instead of hardcoded 512 bytes

heroin-moose:
Fix another couple of tests that fail on 4Kn drives:

* iSCSI.iSCSITMF.AbortTaskSimpleAsync
* iSCSI.iSCSITMF.LUNResetSimpleAsync


bvanassche:
Thanks for the patch!

PR #338, opened 2020-08-18
pdu: fix use after free during cancellation

ddiss:
Fixes: 10868c4 ("libiscsi: Avoid discontinuities in cmdsn ordering in some cases")


bvanassche:
Thanks for the patch!

PR #341, opened 2020-08-31
Assert reconnect succeeded after async TMF completes

drssds:
In the test iSCSI.iSCSITMF.LUNResetSimpleAsync
CU_ASSERT_EQUAL(reconnect_succeeded, 1) must be called after the async
TMF command completes.

To avoid any race conditions, moved this assert to the TMF completion callback.


bvanassche:
This patch has been applied manually on the master branch. Thanks for the patch!

PR #342, opened 2020-09-02
test-tool: Fix CmdSN tests

heroin-moose:
After commit 10868c491d1d16fe2879fdba30a95d855fa936a6 CmdSN tests no
longer work because iscsi->cmdsn is incremented during the second test
phase (sending "good" TUR after "bad" TUR), so the resulting CmdSN is
not ExpCmdSN, but ExpCmdSN + 1 that is not acknowledged by the target.
This commit fixes the issue by setting iscsi->cmdsn to iscsi->expcmdsn - 1.

Affected tests:

 * iSCSI.iSCSIcmdsn.iSCSICmdSnTooLow
 * iSCSI.iSCSIcmdsn.iSCSICmdSnTooHigh


bvanassche:
Thanks for the patch!

PR #343, opened 2020-09-11
test-tool: add version descriptor checks for more SBC-3 versions

almilos:
If device claims SBC-3 version in its version descriptors it won't be marked as SBC-3 compliant.

This commit adds checks for:
* SBC-3 T10/BSR INCITS 514 revision 35
* SBC-3 T10/BSR INCITS 514 revision 36
* SBC-3 ANSI INCITS 514-2014


bvanassche:
Thanks for the patch!

PR #344, opened 2020-10-14
test-tool: check compare and write miscompare sense information

ddiss:


lib/iscsi-command.c
```diff
@@ -327,6 +327,13 @@ static void parse_sense_desc_info(struct scsi_sense *sense,
 	sense->information = scsi_get_uint64(&inf[0]);
 }
 
+static void parse_sense_fixed_info(struct scsi_sense *sense,
```
bvanassche, line 330: Please inline this function. I don't think it's worth it to introduce a new function for these two statements.

lib/iscsi-command.c
```diff
@@ -331,6 +339,11 @@ static void parse_sense_descriptors(struct scsi_sense *sense, const uint8_t *sb,
 		if (!(p[2] & 0x80)) /* VALID bit */
```
bvanassche, line 346: Do you agree that not all descriptors have a VALID bit and hence that this if-statement should be moved inside the switch-statement?

lib/iscsi-command.c
```diff
@@ -319,6 +319,14 @@ static void parse_sense_spec(struct scsi_sense *sense, const uint8_t inf[3])
 	sense->field_pointer = scsi_get_uint16(&inf[1]);
 }
 
+/* Information field length differs between Descriptor and Fixed types */
```
bvanassche, line 322: Since the body of this function is only two lines long, please inline this function.

bvanassche (changes requested):

lib/iscsi-command.c
ddiss: Good catch, it looks like `VALID` isn't actually present for `Sense key specific sense data descriptor`, at least not SPC-5 r17 . I'll split this out as a separate fix.

bvanassche:
Thanks for the patches!

PR #345, opened 2020-10-20
Improve WriteResidual tests

nastya-nizharadze:
Linux target implementation, TCM, doesn’t pass write residuals tests (Write10Residuals, Write12Residualds, Write16Residuals).

One of the reason why they fail is because TCM doesn’t complete writes that have residuals with GOOD status, instead it replies with CHECK CONDITION and INVALID FIELD IN CDB sense code. The sense code doesn’t seem to be a valid reply but it's legit for a target to reply with CHECK CONDITION as outlined in [the discussion on T10](https://www.t10.org/pipermail/t10/2012-September/016350.html), RFC 7143 (11.4.5.1.  Field Semantics): 
```
Targets may set the residual count, and initiators may use
   it when the response code is Command Completed at Target (even if the
   status returned is not GOOD)
```

and FCP-4 (9.4.2 FCP_DATA IUs for read and write operations):
```
If the command requested that data beyond the length specified by the FCP_DL field be transferred, then the device server shall set the FCP_RESID_OVER bit (see 9.5.8) to one in the FCP_RSP IU and:
a) process the command normally except that data beyond the FCP_DL count shall not be requested or transferred;
b) transfer no data and return CHECK CONDITION status with the sense key set to ILLEGAL REQUEST and the additional sense code set to INVALID FIELD IN COMMAND INFORMATION UNIT; or
c) may transfer data and return CHECK CONDITION status with the sense key set to ABORTED COMMAND and the additional sense code set to INVALID FIELD IN COMMAND INFORMATION UNIT.
```

To allow different replies from different targets, b) from FCP-4 was added as an alternative valid reply. With the PR, libiscsi supports both GOOD and CHECK CONDITION with INVALID FIELD IN COMMAND INFORMATION UNIT.

The tests have also been refactored to make sure that no transfer happens if target replies with INVALID FIELD IN COMMAND INFORMATION UNIT. As a side effect of the refactoring, most of the code duplication was removed.

After refactoring, it turned out that some test cases duplicate each other, so they were removed.

As soon as I send patches in TCM to fix sense key I will attach the link to it here.

### DIFF
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -37,6 +37,9 @@ extern "C" {
 #ifndef MAX
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 #endif
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+#endif
 
 #define ISCSI_RAW_HEADER_SIZE			48
 #define ISCSI_DIGEST_SIZE			 4
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -40,6 +40,9 @@ extern "C" {
 #ifndef ARRAY_SIZE
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
 #endif
+#ifndef DIV_ROUND_UP
+#define DIV_ROUND_UP(x, y) (((x) + (y) - 1) / (y))
+#endif
 
 #define ISCSI_RAW_HEADER_SIZE			48
 #define ISCSI_DIGEST_SIZE			 4
--- a/test-tool/test_write10_residuals.c
+++ b/test-tool/test_write10_residuals.c
@@ -25,21 +25,34 @@
 #include "iscsi-private.h"
 #include "scsi-lowlevel.h"
 #include "iscsi-test-cu.h"
-
-static int check_status (struct scsi_task *task) {
-        return (task->status     == SCSI_STATUS_GOOD ||
-               (task->status     == SCSI_STATUS_CHECK_CONDITION &&
-                task->sense.key  == SCSI_SENSE_ILLEGAL_REQUEST &&
-                task->sense.ascq == SCSI_SENSE_ASCQ_INVALID_FIELD_IN_INFORMATION_UNIT));
-}
+#include "test_write_residuals.h"
 
 void
 test_write10_residuals(void)
 {
-        struct scsi_task *task_ret;
-        unsigned char buf[10000];
-        struct iscsi_data data;
-        unsigned int i;
+        /* testing scenarios */
+        const struct residuals_test_data write10_residuals[] = {
+        /* cdb_size, xfer_len,        buf_len,          residuals_kind,   residuals_amount, check_overwrite */
+                {10,        1,              0,  SCSI_RESIDUAL_OVERFLOW,         block_size, false,
+                 "Try writing one block but with iSCSI expected transfer length==0"},
+
+                {10,        1,          10000, SCSI_RESIDUAL_UNDERFLOW, 10000 - block_size, false,
+                 "Try writing one block but with iSCSI expected transfer length==10000"},
+
+                {10,        1,            200,  SCSI_RESIDUAL_OVERFLOW,   block_size - 200, false,
+                 "Try writing one block but with iSCSI expected transfer length==200"},
+
+                {10,        2,     block_size,  SCSI_RESIDUAL_OVERFLOW,         block_size, false,
+                 "Try writing two blocks but iSCSI expected one block transfer length"},
+
+                {10,        1, 2 * block_size, SCSI_RESIDUAL_UNDERFLOW,         block_size,  true,
+                 "Verify that if iSCSI EDTL > SCSI TL then we only write SCSI TL amount of data"},
+
+                {10,        2,     block_size,  SCSI_RESIDUAL_OVERFLOW,         block_size,  true,
+                 "Verify that if iSCSI EDTL < SCSI TL then we only write iSCSI EDTL amount of data"},
+        };
+
+        unsigned int i = 0;
 
         logging(LOG_VERBOSE, LOG_BLANK_LINE);
         logging(LOG_VERBOSE, "Test WRITE10 commands with residuals");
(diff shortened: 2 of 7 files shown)


roolebo (approved):
It was thoroughly reviewed internally and tested with the changes to be sent upstream: https://github.com/roolebo/linux/commits/target-write-residuals-v1-feedback

test-tool/test_write10_residuals.c
```diff
@@ -75,11 +75,15 @@ test_write10_residuals(void)
                 return;
         }        
         logging(LOG_VERBOSE, "Verify that the target returned SUCCESS");
-        if (task->status != SCSI_STATUS_GOOD) {
+        ok = task->status == SCSI_STATUS_GOOD ||
```
bvanassche: This is wrong. A SCSI target must not report "ILLEGAL REQUEST" in case of a residual.

bvanassche (changes requested):

test-tool/test_write10_residuals.c
nastya-nizharadze: Hi @bvanassche,

It is a good practice to separate a refactoring and a changes for the more convenient review.  All repetitions are removed in the next commit. Even if I introduce a function or a macro, it will be removed by the next commit. However, if you think that it is really better to remove the duplication in this commit, please tell me and I will do it.

Thanks!

nastya-nizharadze:
Hi @bvanassche,

Our patches have been added to the [5.12/scsi-staging](https://git.kernel.org/pub/scm/linux/kernel/git/mkp/scsi.git/log/?h=5.12/scsi-staging)

Now TCM send ASCQ = 0x03, INVALID FIELD IN COMMAND INFORMATION UNIT for residual write case.

Could you please tell me whether it is really necessary to make the previously requested changes?
Also if there is something else I could do/improve as part of these changes?

Thanks!

test-tool/test_write_residuals.h
```diff
@@ -0,0 +1,48 @@
...
+
+
+struct residuals_test_data {
+  size_t cdb_size;
+  unsigned int xfer_len;
+  unsigned int buf_len;
+  unsigned int residuals_kind;
+  size_t residuals_amount;
+  bool check_overwrite;
+  const char log_messages[100];
```
bvanassche: Please change the type of the log message from const char [100] into const char* to reduce the space occupied by this data structure.

test-tool/test_write_residuals.h
```diff
@@ -0,0 +1,48 @@
...
+};
+
+extern bool command_is_implemented;
```
bvanassche, line 47: What is the purpose of this global variable? Since it is set but not used, can it be removed?

test-tool/test_write_residuals.h
```diff
@@ -0,0 +1,48 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <stdbool.h>
+
+#include <CUnit/CUnit.h>
+
+#include "iscsi.h"
+#include "iscsi-private.h"
+#include "scsi-lowlevel.h"
+#include "iscsi-test-cu.h"
```
bvanassche, line 31: Please remove all include directives except #include <stddef.h> (for size_t).

test-tool/test_write10_residuals.c
```diff
@@ -25,16 +25,34 @@
...
 void
 test_write10_residuals(void)
 {
-        struct scsi_task *task_ret;
-        unsigned char buf[10000];
-        struct iscsi_data data;
-        int ok;
-        unsigned int i;
+        /* testing scenarios */
+        const struct residuals_test_data write10_residuals[] = {
```
bvanassche, line 34: Please change 'const' into 'static const'.

test-tool/test_write12_residuals.c
```diff
@@ -25,16 +25,34 @@
...
 void
 test_write12_residuals(void)
 {
-        struct scsi_task *task_ret;
-        unsigned char buf[10000];
-        struct iscsi_data data;
-        int ok;
-        unsigned int i;
+        /* testing scenarios */
+        const struct residuals_test_data write12_residuals[] = {
```
bvanassche, line 34: Same comment here and in the other source files: please change 'const' into 'static const'.

bvanassche (changes requested):

test-tool/test_write_residuals.c
```diff
@@ -158,6 +162,26 @@ write_residuals_test(const struct residuals_test_data *tdata)
...
 
+        /* According to FCP target could transfer no data and return 
+           CHECK CONDITION status with the sense key set to 
+           ILLEGAL REQUEST and the additional sense code set to 
+           INVALID FIELD IN COMMAND INFORMATION UNIT; this check prevent
+           false assert*/
+
+        if (scsi_status != SCSI_STATUS_GOOD) {
+                logging(LOG_VERBOSE, "Verify that blocks were NOT "
+                        "overwritten and still contains 'a'");
```
bvanassche: contains -> contain?

bvanassche (changes requested):

include/iscsi-private.h
```diff
@@ -40,6 +40,9 @@ extern "C" {
 #ifndef ARRAY_SIZE
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
 #endif
+#ifndef DIV_ROUND_UP
+#define DIV_ROUND_UP(x, y) (((x) + ((y) - 1)) / (y))
```
bvanassche, line 44: There are more parentheses in this expression than necessary. I think that ((y) - 1) can be changed into (y) - 1 without changing the result of this expression.

bvanassche (changes requested):

test-tool/test_write10_residuals.c
bvanassche: It is a good practice not to introduce repetitive code. Please introduce a helper function in this commit that returns the result of the following expression:

task->status == SCSI_STATUS_GOOD ||
￼                (task->status == SCSI_STATUS_CHECK_CONDITION &&
￼                 task->sense.key == SCSI_SENSE_ILLEGAL_REQUEST &&
￼                 task->sense.ascq == SCSI_SENSE_ASCQ_INVALID_FIELD_IN_INFORMATION_UNIT

bvanassche (changes requested):

test-tool/test_write10_residuals.c
nastya-nizharadze: done

test-tool/test_write_residuals.h
nastya-nizharadze: This variable is needed because if the command is not implemented, testing should end. It is used in test_write10_residuals.c, test_write12_residuals.c and test_write12_residuals.c
```suggestion
                if (!command_is_implemented) {
                        CU_PASS("WRITE10 is not implemented.");
                        return;
                }
```

test-tool/test_write_residuals.h
nastya-nizharadze: done

test-tool/test_write10_residuals.c
nastya-nizharadze: I am using `block_size` variable to initialize this structure. Therefore I can not make it static.

test-tool/test_write_residuals.c
nastya-nizharadze: done

include/iscsi-private.h
nastya-nizharadze: done

test-tool/test_write_residuals.c
```diff
@@ -82,9 +83,9 @@ write_residuals_test(const struct residuals_test_data *tdata)
         memset(task, 0, sizeof(*task));
 
         if (tdata->check_overwrite) {
-                memset(scratch, 'b', (tdata->xfer_len * tdata->buf_len));
+                memset(scratch, 'b', tdata->buf_len);
         } else {
-                memset(scratch, 0xa6, (tdata->xfer_len * tdata->buf_len));
+                memset(scratch, 0xa6, tdata->buf_len);
```
bvanassche: Are these changes perhaps a bug fix for the previous patch in this series? If so, please move these into the previous patch.

test-tool/test_write_residuals.h
```diff
@@ -0,0 +1,37 @@
...
+#include <stdbool.h>
+
+struct residuals_test_data {
+  size_t cdb_size;
+  unsigned int xfer_len;
+  unsigned int buf_len;
+  unsigned int residuals_kind;
+  size_t residuals_amount;
+  bool check_overwrite;
+  const char *log_messages;
```
bvanassche: Please document the meaning of each member of this data structure.

test-tool/test_write_residuals.c
nastya-nizharadze: done

test-tool/test_write_residuals.h
nastya-nizharadze: done

nastya-nizharadze:
Thanks!

PR #347, opened 2020-11-08
Fix iser sigsegv v2

puhou:
Fix segmentation fault when using ISER. See commit message for details.


bvanassche:
Thanks for the patches!

PR #350, opened 2021-01-24
test-tool: test various ALLOC LENGTHS for REPORT LUNS

ddiss:
SPC-3 6.21 REPORT LUNS command states:
  Device servers compliant with SPC return CHECK CONDITION status, with
  the sense key set to ILLEGAL REQUEST, and the additional sense code
  set to INVALID FIELD IN CDB when the allocation length is less than
  16 bytes.

Test this requirement by sending REPORT LUNS requests with ALLOCATION
LENGTH values under, equal to and over 16 bytes.

### DIFF
--- a/test-tool/test_report_luns.c
+++ b/test-tool/test_report_luns.c
@@ -19,6 +19,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdlib.h>
+#include <poll.h>
 
 #include <CUnit/CUnit.h>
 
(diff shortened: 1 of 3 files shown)


test-tool/test_report_luns.c
```diff
@@ -64,3 +65,90 @@ test_report_luns_simple(void)
...
+
+	logging(LOG_VERBOSE, LOG_BLANK_LINE);
+	logging(LOG_VERBOSE, "Test REPORT LUNS - ALLOCATION LENGTH");
+
+	/*
+	 * SPC-3 6.21 REPORT LUNS command:
+	 * NOTE 30 - Device servers compliant with SPC return CHECK CONDITION
+	 * status, with the sense key set to ILLEGAL REQUEST, and the additional
+	 * sense code set to INVALID FIELD IN CDB when the allocation length is
+	 * less than 16 bytes.
```
ddiss, line 118: I noticed that the SPC-4 REPORT LUNS spec doesn't include this specific note regarding allocation length < 16. Instead it refers to the generic ALLOCATION LENGTH field is definition (in 4.2.5.6), which carries:
```
An allocation length of zero specifies that no data shall be transferred. This condition shall not be considered
an error.
...
If the amount of information that is available to be transferred exceeds the maximum value that the
ALLOCATION LENGTH field in combination with other fields in the CDB is capable of specifying, then no data
shall be transferred and the command shall be terminated with CHECK CONDITION status, with the sense
key set to ILLEGAL REQUEST, and the additional sense code set to INVALID FIELD IN CDB.
```

The first clause above probably means that the `i = 0` case below is invalid with SPC-4. @bvanassche what do you think, should I drop this whole test, or change `i = 0` initialization to `i = 1`. I think the `i =1` change should probably satisfy both SPC-3 and SPC-4 REPORT LUNS specifications by ignoring ALLOCATION LENGTH = 0 behaviour.

bvanassche: I propose to ignore note 30 from SPC-3 since that clause seems inconsistent to me with how the allocation length field of other SCSI commands is handled. Additionally, I haven't found that clause in SPC-4, SPC-5 nor SPC-6.

PR #352, opened 2021-05-23
Makefile.am improvements.

orbea:
These are some improvements for the `Makefile.am` files.

1. When linking dependencies for the whole `Makefile.am` the variable is `LIBS`, not `LDADD`. This removes `LDADD` and replaces it with `LIBS` as needed.
2. When linking dependencies the correct variable is `libfoo_LIBADD` for libraries and `foo_LDADD` for programs while `LDFLAGS` should only be used for other linker flags such as `-no-undefined` or `-version-info`.
3.  This adds `-no-undefined` where applicable to help detect the undefined references fixed in the first commit. Please note that GNU libtool has a long standing design where `-no-undefined` is silently ignored and will not be respected. However when using slibtool (https://dev.midipix.org/cross/slibtool) this will not be a problem and `-no-undefined` works as expected.


bvanassche:
Thanks for the patches!

PR #355, opened 2021-06-17
Inquiry for Block Device Charactertics expected to succeed always

nvenkataddn:
In the iscsi-test-cu.c, inquiry is being issued with block device characteristics page and if the command fails with any error other than invalid opcode, it causes a test failure. However, block device characteristics page is an optional command and the target may not be supporting this inquiry. So as a fix moved the test out of the main test code and made it part of the inquiry test suites.

### DIFF
--- a/test-tool/iscsi-test-cu.c
+++ b/test-tool/iscsi-test-cu.c
@@ -87,6 +87,7 @@ static CU_TestInfo tests_inquiry[] = {
         { "AllocLength", test_inquiry_alloc_length},
         { "EVPD", test_inquiry_evpd},
         { "BlockLimits", test_inquiry_block_limits},
+        { "BlockDeviceCharacteristics", test_inquiry_block_device_characteristics},
         { "MandatoryVPDSBC", test_inquiry_mandatory_vpd_sbc},
         { "SupportedVPD", test_inquiry_supported_vpd},
         { "VersionDescriptors", test_inquiry_version_descriptors},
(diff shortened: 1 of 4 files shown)


test-tool/iscsi-test-cu.c
```diff
@@ -1404,19 +1405,6 @@ main(int argc, char *argv[])
                 }
         }
 
-        /* try reading block device characteristics vpd */
```
bvanassche, line 1407: Converting this code into a unit test seems wrong to me because several tests need the block device characteristics page contents (test-tool/test_sanitize_block_erase.c, test-tool/test_sanitize_crypto_erase.c and test-tool/test_sanitize_overwrite.c) and converting this code into a unit test creates a dependency between tests that is hard to track.

nvenkataddn: I feel that such tests should first check if the BDC page is valid before running the test. Block Device Characteristics page is not mandatory and when checking for compliance, it should not be a mandatory test for the compliance suite to work.

bvanassche:
Regarding the following code:
```
        /* try reading block device characteristics vpd */
        inquiry(sd, &inq_bdc_task, 1, SCSI_INQUIRY_PAGECODE_BLOCK_DEVICE_CHARACTERISTICS, 255,
                EXPECT_STATUS_GOOD);
        if (inq_bdc_task == NULL || inq_bdc_task->status != SCSI_STATUS_GOOD) {
                printf("Failed to read Block Device Characteristics page\n");
        } else {
                inq_bdc = scsi_datain_unmarshall(inq_bdc_task);
                if (inq_bdc == NULL) {
                        printf("failed to unmarshall inquiry datain blob\n");
                        goto err_sds_free;
                }
        }
```
Please help me to understand what is wrong with the above code. My understanding of the above code is that it prints an error message if the INQUIRY command fails and also that it only exits if the BDC data is invalid?

bvanassche:
Please reopen after the review comments have been addressed.

PR #359, opened 2021-07-29
test: fix ExtendedCopy.Large SKIPPED cases

ddiss:
CU_PASS() needs to be coupled with a return to actually skip the test.
Similarly, the CU_PASS message isn't printed, so a logging() call is
also needed here.


bvanassche:
Thanks for the patch!

PR #360, opened 2021-08-19
Add missing include assert.h library

wawryk:


bvanassche:
Thanks for the patch!

PR #362, opened 2021-08-20
Examples various fixes

srowe:
Minor changes to output text in iscsi-dd


bvanassche:
Thanks for the patches!

PR #364, opened 2021-09-12
Added some necessary files into the dist tarball for rpm making

tianrenz2:
Hi, I got some problems when making rpm based on the current master branch, hope the following changes would help.


bvanassche:
Thanks for the patches!

PR #365, opened 2021-10-01
Two little fixes.

x2c3z4:
Patch 1:
Avoid to include the private file;

Patch 2:
Fix the memory leak.


bvanassche:
Patch 2 has been applied but patch 1 not since I'm not sure it's an improvement.

PR #366, opened 2021-10-01
New test for Third-party Copy VPD page

Ser01x:
This test covers the Third-party Copy VPD page.

In this step, we check only mandatory Third-party copy descriptors for block devices. Its are:

- Supported Commands
- Parameter Data
- Supported Descriptors
- Supported CSCD Descriptor IDs
- General Copy Operations

### DIFF
--- a/include/scsi-lowlevel.h
+++ b/include/scsi-lowlevel.h
@@ -689,30 +689,84 @@ struct scsi_inquiry_device_identification {
 	struct scsi_inquiry_device_designator *designators;
 };
 
+/* Third-party copy descriptor type codes */
 enum third_party_copy_descriptor_type {
-	THIRD_PARTY_COPY_TYPE_SUPPORTED_COMMANDS  = 0x0001
+	THIRD_PARTY_COPY_TYPE_SUPPORTED_COMMANDS            = 0x0001,
+	THIRD_PARTY_COPY_TYPE_PARAMETER_DATA                = 0x0004,
+	THIRD_PARTY_COPY_TYPE_SUPPORTED_DESCRIPTORS         = 0x0008,
+	THIRD_PARTY_COPY_TYPE_SUPPORTED_CSCD_DESCRIPTORS_ID = 0x000c,
+	THIRD_PARTY_COPY_TYPE_GENERAL_COPY_OPERATIONS       = 0x8001
 };
 
+/* Command support descriptor */
 struct third_party_copy_command_support {
 	struct third_party_copy_command_support *next;
 
 	int operation_code;
-	int service_action_length;
-	int *service_action;
+	int service_actions_list_length;
+	int *service_actions;
 };
 
+/* Supported Commands third-party copy descriptor */
 struct third_party_copy_supported_commands {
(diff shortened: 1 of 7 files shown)


Ser01x:
Hi Bart,

It is the first version for review. I'm not sure. Do we need to validate a Third-party VPD page so in-depth?

lib/scsi-lowlevel.c
```diff
@@ -1487,11 +1487,6 @@ scsi_inquiry_unmarshall_device_identification(struct scsi_task *task)
 	return inq;
```
bvanassche, line 1487: Please use the imperative mood for patch descriptions. That means changing `cleaned a dead codes` into e.g. `Remove dead code`. Otherwise this patch looks good to me.

include/scsi-lowlevel.h
```diff
@@ -689,30 +689,84 @@ struct scsi_inquiry_device_identification {
 	struct scsi_inquiry_device_designator *designators;
```
bvanassche, line 689: In the patch description, please change `Its` into `These`.

include/scsi-lowlevel.h
```diff
@@ -689,30 +689,84 @@ struct scsi_inquiry_device_identification {
...
+/* Parameter Data third-party copy descriptor */
+struct third_party_copy_parameter_data {
+	enum third_party_copy_descriptor_type descriptor_type;
+	int descriptor_length;
+	unsigned char reserved_1[4];
+	int maximum_cscd_descriptor_count;
+	int maximum_segment_descriptor_count;
+	unsigned int maximum_descriptor_list_length;
+	unsigned int maximum_inline_data_length;
+	unsigned char reserved_2[12];
```
bvanassche, line 727: Does this data structure come from a SCSI standard? If so, please use uint8_t, uint8_t, uint32_t etc. instead of int / unsigned char. If not, please remove the "reserved" members.

include/scsi-lowlevel.h
```diff
@@ -689,30 +689,84 @@ struct scsi_inquiry_device_identification {
...
+/* General Copy Operations third-party copy descriptor */
+struct third_party_copy_general_copy_operations {
+	enum third_party_copy_descriptor_type descriptor_type;
+	int descriptor_length;
+	unsigned int total_concurrent_copies;
+	unsigned int maximum_identified_concurrent_copies;
+	unsigned int maximum_segment_length;
+	int data_segment_granularity;
+	int inline_data_granularity;
+	unsigned char reserved[18];
```
bvanassche, line 755: What is the role of the `reserved` member in this data structure?

bvanassche (changes requested):

bvanassche:
> Hi Bart,
> 
> It is the first version for review. I'm not sure. Do we need to validate a Third-party VPD page so in-depth?

I don't have a strong opinion about how detailed the tests should be. I will leave it up to you to decide how detailed you want the tests to be in the first version of this test :-)

Ser01x:
I need to rebase my copy of this repository for prepare the v2 patches.

PR #367, opened 2022-01-17
add libiscsi.syms to .gitignore

jlevon:
add libiscsi.syms to .gitignore


bvanassche:
Thanks for the patch!

PR #368, opened 2022-01-17
add iscsi_force_reconnect()

jlevon:
If a connection attempt is hung, then iscsi_reconnect() won't do anything. This
makes sense if we'd just re-try to connect to the same target, but if (for
example) login redirect might point us to a different, healthy, target, it
should be possible to restart the full connection process on request.

Signed-off-by: John Levon <john.levon@nutanix.com>


bvanassche:
Thanks for the patch!

PR #369, opened 2022-02-11
iscsi-pr: add persistent reservation tool

pizhenwei:
iscsi-pr uses libiscsi to connect target and issues PR command.
In most cases, iscsi-pr use the same parameters as sg_persist(from sg3-utils).

For example, use iscsi-pr to dump keys by command:
  ~# ./utils/iscsi-pr --read-keys iscsi://192.168.122.44/iqn.2003-01.org.linux-iscsi.bytedance.x8664:sn.a6b7bff3d509/0 -i iqn.2005-03.org.open-iscsi:12345678
  PR generation=0x20, 19 registered reservation keys follow:
    0xabcd1234
    0xabcd
    0xabcd
    0xabc

A few command is not implemented in this patch, add them if necessary
in future.

Signed-off-by: zhenwei pi <pizhenwei@bytedance.com>

### DIFF
--- a/utils/iscsi-pr.c
+++ b/utils/iscsi-pr.c
@@ -1,0 +1,485 @@
+/* 
+   Copyright (C) 2022 by zhenwei pi<pizhenwei@bytedance.com>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <getopt.h>
+#include <inttypes.h>
+#include "iscsi.h"
+#include "scsi-lowlevel.h"
+
+static void pr_in_read_keys(struct scsi_persistent_reserve_in_read_keys *rk)
+{
+        int i;
+
+        printf("  PR generation=0x%x, %d registered reservation keys follow:\n",
+                        rk->prgeneration, rk->num_keys);
+        for (i = 0; i < rk->num_keys; i++) {
+                printf("    0x%" PRIx64 "\n", rk->keys[i]);
+        }
+}
+
+static void pr_in_read_reservation(struct scsi_persistent_reserve_in_read_reservation *rr)
+{
+        printf("  PR generation=0x%x, Reservation follows:\n", rr->prgeneration);
+        if (!rr->reserved) {
+                return;
+        }
+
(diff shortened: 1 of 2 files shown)


utils/iscsi-pr.c
```diff
@@ -0,0 +1,476 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <getopt.h>
+#include <inttypes.h>
+#include "iscsi.h"
+#include "scsi-lowlevel.h"
+
+static const char *initiator = "iqn.2022.com.github:pizhenwei:libiscsi:iscsi-pr";
```
bvanassche, line 35: Please do not hardcode iSCSI IQNs.

utils/iscsi-pr.c
```diff
@@ -0,0 +1,476 @@
...
+
+static void pr_in_read_keys(struct scsi_persistent_reserve_in_read_keys *rk)
+{
+	int i;
```
bvanassche, line 39: Please use the same coding style as the rest of the libiscsi code: indentation level 8, no tabs.

utils/iscsi-pr.c
```diff
@@ -0,0 +1,476 @@
...
+		printf("    scope: %d,  type: %s\n",
+				rr->pr_scope, scsi_pr_type_str(rr->pr_type));
+	}
+}
+
+static void pr_in_report_capabilities(struct scsi_persistent_reserve_in_report_capabilities *rc)
+{
+	uint8_t  i;
+	uint16_t __mask = rc->persistent_reservation_type_mask;
+	uint16_t mask = ((__mask & 0xff) << 8) | (__mask >> 8);
```
bvanassche, line 68: Please use scsi_get_uint*() for conversion from big endian to little endian.

bvanassche (changes requested):

utils/iscsi-pr.c
pizhenwei: It's possible to overwrite the default initiator by `-i, --initiator-name=iqn-name`, keep the same parameter with other iscsi utils. (Ex, iscsi-inq)

utils/iscsi-pr.c
pizhenwei: @bvanassche Hi, could you give me any hint about above questions? Or I just format code without tabs and force push a new version?

utils/iscsi-pr.c
```diff
@@ -0,0 +1,491 @@
...
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef HAVE_CONFIG_H
```
bvanassche, line 18: Can this code block be left out? I don't see any poll() calls so why has <poll.h> been included? And if #include <poll.h> is left out, why to include <config.h>?

bvanassche (changes requested):

utils/iscsi-pr.c
pizhenwei: Yes, removed this block and other 2 header files, and force pushed a new version.
Thanks a lot!

bvanassche (approved):

PR #371, opened 2022-03-02
iscsi-command: Fix leak in iscsi_send_data_out

raphael-nutanix:
In iscsi_send_data_out() a PDU is allocated, but there is no error
handling logic to free it if the PDU cannot be queued.
iscsi_allocate_pdu() may allocate memory for the PDU. This memory may be
leaked if iscsi_queue_pdu() fails since there is no call to free it.

Orignally there was a free() call but it was removed as a part of a
cleanup adding checks for NULL pdu callbacks.

Fixes: 423b82efa4bd ("pdu: check callback for NULL everywhere")
Reviewed-by: Manish Mishra <manish.mishra@nutanix.com>
Signed-off-by: Raphael Norwitz <raphael.norwitz@nutanix.com>


bvanassche:
Thanks for the patch!

PR #375, opened 2022-04-06
iscsi-inq: handle setting of debug_level correctly

wenchao-hao:
According to the man page and help info, --debug=integer can specify the
debug_level, while it would report following error:

iscsi-inq --debug=2
iscsi-inq: option '--debug' doesn't allow an argument

It's because the iscsi-inq code did not handle this parameters
correctly. So here we just correct it.

Signed-off-by: Wenchao Hao <haowenchao@huawei.com>


wenchao-hao:
@bvanassche

bvanassche:
Hi Wenchao, mentioning me explicitly in pull requests or bugs is not necessary. I receive an email every time there is activity on the libiscsi project. Your patch looks good to me. I see that Ronnie Sahlberg already merged your patch.

PR #378, opened 2022-06-01
check whether malloc success

sallyjunjun:
Add check after malloc to avoid dereferencing null return value


lib/sync.c
```diff
@@ -1868,6 +1868,9 @@ iscsi_discovery_cb(struct iscsi_context *iscsi, int status,
                 struct iscsi_discovery_address *datmp;
 
                 datmp = malloc(sizeof(struct iscsi_discovery_address));
+		if (datmp == NULL) {
+			return;
+		}
```
bvanassche, line 1873: Please follow the existing code style and use spaces to indent code instead of tabs.

PR #379, opened 2022-06-08
fix iscsi-ls parameter parse

sallyjunjun:
If invalid option is input with iscsi-ls，such as "iscsi-ls -a iscsi://", the command just stuck here and do not print useful information for the user to correct.
Fix this problem with getopt_long.


bvanassche:
Thanks for the patch!

PR #381, opened 2022-06-08
fix dereferencing NULL return value

sallyjunjun:
Check return value of scsi_malloc to avoid dereferencing NULL return value.


bvanassche:
Thanks for the patch!

PR #383, opened 2022-06-15
Fix segmentation fault problem.

sallyjunjun:
When execute iscsi_task_mgmt_lun_reset_async function,
pdus are already removed from waitpdu list. In iscsi_service
function, this will call iscsi_process_pdu and release
pdu from waitpdu again, which cause segmentation fault.

Whether waitpud list is NULL should be checked here to avoid
the problem.

Signed-off-by: geruijun <geruijun@huawei.com>


bvanassche:
Thanks for the patch!

PR #389, opened 2022-12-14
aros_compat: Add check after malloc allocation

likunyur:
Increase the check after malloc allocation, which should be helpful to the robustness of the program.


lib/connect.c
```diff
@@ -469,6 +469,9 @@ static int reconnect(struct iscsi_context *iscsi, int force)
 		tmp_iscsi->old_iscsi = iscsi->old_iscsi;
 	} else {
 		tmp_iscsi->old_iscsi = malloc(sizeof(struct iscsi_context));
+		if (!tmp_iscsi->old_iscsi)
```
sahlberg, line 472: need to free the tmp_iscsi context too

sahlberg (changes requested):

PR #394, opened 2023-07-07
utils: use strtol instead of atoi

pizhenwei:
HEX format is friendly to iSCSI utils, for example: ./iscsi-inq -e 0x1 -c 0xb1 iscsi://...

atoi supports decimal only, this example does not work. Use strtol(nptr, NULL, 0) to auto-detect format, then this works fine.


bvanassche:
Thanks for the patch!

PR #397, opened 2023-08-08
iscsi-discard: adapt block limits

pizhenwei:
The blocks may exceed the block limits offered by the target, query unmap/write zero max blocks by inquiry
SCSI_INQUIRY_PAGECODE_BLOCK_LIMITS command, than use a loop to do the job.


pizhenwei:
Hi @sahlberg 
PING

Issue #409, opened 2024-01-30
segfault


folkertvanheusden:
With a iSCSI-target which not yet implements any mode sense pages, there's a segfault in set_swp as `ms->pages` is then never set.

```
Suite: ModeSense6
  Test: AllPages ...
    Test of MODESENSE6 AllPages
    Send MODESENSE6 command to fetch AllPages
    Send MODESENSE6 (Expecting SUCCESS) 
    [OK] MODESENSE6 returned SUCCESS NO SENSE(0x00) UNKNOWN(0x0000)
    [SUCCESS] All Pages fetched.
    Try to unmarshall the DATA-IN buffer.
    [SUCCESS] Unmarshalling successful.
    Verify that mode data length is >= 3
    [FAILED] Mode data length is < 3
FAILED
    1. test_modesense6_all_pages.c:61  - CU_ASSERT_TRUE(ms->mode_data_length >= 3)
  Test: Control ...
    Test of MODESENSE6 CONTROL page
    Fetch the CONTROL page via AllPages
    Send MODESENSE6 command to fetch AllPages
    Send MODESENSE6 (Expecting SUCCESS) 
    [OK] MODESENSE6 returned SUCCESS NO SENSE(0x00) UNKNOWN(0x0000)
    [SUCCESS] All Pages fetched.
    Try to unmarshall the DATA-IN buffer.
    [SUCCESS] Unmarshalling successful.
    Verify that mode data length is >= 3
    [FAILED] Mode data length is < 3
    [WARNING] CONTROL page was not returned from AllPages. All devices SHOULD implement this page.
    Fetch the CONTROL page directly
    Send MODESENSE6 command to fetch CONTROL
    Send MODESENSE6 (Expecting SUCCESS) 
    [OK] MODESENSE6 returned SUCCESS NO SENSE(0x00) UNKNOWN(0x0000)
    [SUCCESS] CONTROL page fetched.
    Try to unmarshall the DATA-IN buffer.
    [SUCCESS] Unmarshalling successful.
    Verify that mode data length is >= 3
    [FAILED] Mode data length is < 3
    [WARNING] CONTROL page was not returned.All devices SHOULD implement this page.
    [SKIPPED] CONTROL page is not implemented.
FAILED
    1. test_modesense6_control.c:65  - CU_ASSERT_TRUE(ms->mode_data_length >= 3)
    2. test_modesense6_control.c:103  - CU_ASSERT_TRUE(ms->mode_data_length >= 3)
[...]

PR #410, opened 2024-01-30
Fix for https://github.com/sahlberg/libiscsi/issues/409

folkertvanheusden:
'ms->pages' was not checked for being NULL. This can happen when a target does not return any pages.


bvanassche:
Thanks for the patch!

Issue #415, opened 2024-02-13
iscsi-test-cu & logging


folkertvanheusden:
Hi,

Would you accept a patch that adds (in verbose mode) context related debug-logging?
E.g. when a certain iSCSI command fails at a target, that e.g. the ITT (initial task tag) is logged as well.
That would help debugging the target and/or relating problems to network-traces.

PR #416, opened 2024-02-13
A possible 'fix' for issue #415

folkertvanheusden:
This patch adds a timestamp before each logged line. That could help correlating a logging to a network-trace. Because of offsets in time between the tracer and the test and the DUT, this does not always help.


sahlberg:
Looks useful.
Merged, thanks!

PR #418, opened 2024-04-09
allow choosing between libgcrypt, gnutls or embedded MD5

bonzini:
libgcrypt is a relatively large dependency that is used only for the sake of computing MD5 in the CHAP authentication protocol. Some distributions may want to disable it forcibly and rely on the embedded MD5 implementation.

On the other hand, other distributions may want to standardize on as few crypto libraries as possible, and to that end it makes sense to support libgnutls which has a superset of the libgcrypt functionality. So add support for that one as well.


sahlberg:
Merged. Thanks!

Issue #423, opened 2024-09-20
test-tool is not build


folkertvanheusden:
When I run `./configure --enable-test-tool --enable-tests --enable-examples` (and make), the `iscsi-test` is not build (nor is `iscsi-test-cu` build).

folkertvanheusden:
Ah, libcunit1-dev was not installed. that is not mentioned in the docs?

sahlberg:
Doc has been updated.  Thanks!

PR #424, opened 2024-09-20
libcunit1-dev is required for iscsi-test-cu to be build

folkertvanheusden:
See https://github.com/sahlberg/libiscsi/issues/423


sahlberg:
Merged, thanks

PR #427, opened 2024-10-18
For failing writes, show what is in the block instead

folkertvanheusden:
In write_residuals_test "Blocks were overwritten and no longer contain 'a'" is printed, it is helpful when it also print what was in the block instead.


sahlberg:
Merged, thanks!

Issue #428, opened 2024-10-29
Could we support resetting next_reconnect or provide a way to force reconnection?


RayWHL:
Currently, [next_reconnect](https://github.com/sahlberg/libiscsi/blob/master/include/iscsi-private.h#L179) indicates the next available time for reconnection, and its value increases as reconnection attempts fail, potentially reaching up to [30 seconds](https://github.com/sahlberg/libiscsi/blob/master/lib/connect.c#L292). However, if the client detects target recovery during this backoff period and tries to actively reconnect by calling the `iscsi_reconnect` or `iscsi_force_reconnect` interfaces, it cannot trigger a reconnection, because both interfaces are restricted by next_reconnect. If the time hasn’t reached next_reconnect, these interfaces won’t perform any action.

```
	if (time(NULL) < iscsi->next_reconnect) {
		iscsi->pending_reconnect = 1;
		return 0;
	}
```

To address this issue, could we provide a method to reset next_reconnect or a forced reconnection method that bypasses the next_reconnect check to allow the client to proactively reconnect more quickly?

PR #429, opened 2024-10-30
feat: add iscsi_reset_next_reconnect interface

raywang-dev:
When an iSCSI connection enters the reconnection phase, the backoff time (next_reconnect) increases with reconnection retry_cnt. However, if the client detects that the target has recovered before reaching next_reconnect, calling `iscsi_reconnect/iscsi_force_reconnect` has no any effect, making fast reconnection impossible.

This patch introduces an interface to reset next_reconnect, so that the client can reset the backoff time upon detecting target recovery and achieve faster reconnection.

Resolves: https://github.com/sahlberg/libiscsi/issues/428


sahlberg:
Merged, thanks!

PR #433, opened 2024-11-27
lib: finish event_loop on invalid fd

tianrenz2:
When iscsi->fd gets invalid, there is not point to keep stuck in the event loop, instead could give an accurate error about the invalid fd.


sahlberg:
Merged, thanks!

PR #439, opened 2025-03-25
Add some logging improvements

plieven:
+ include LUN into log messages
+ restore logging of local ip and port
+ log resolved ip addresses if portal is a hostname

### DIFF
--- a/lib/logging.c
+++ b/lib/logging.c
@@ -72,9 +72,9 @@ iscsi_log_message(struct iscsi_context *iscsi, int level, const char *format, ..
 	}
 
 	if (iscsi->target_name[0]) {
-		static char message2[1282];
+		static char message2[1294];
 
-		snprintf(message2, 1282, "%s [%s]", message, iscsi->target_name);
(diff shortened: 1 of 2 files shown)


sahlberg:
Merged, thanks.

If you are interested,
I am also working on finally add multithread support to libiscsi.
The bare bones is in the multithreading branch right now.
It still needs to protect all relevant variables/structure fields that could be accessed concurrently but it is work in progress,

If you have ideas  or specific requirements you would like to see, let me know.

This is very similar/based on the multithreading in libnfs and will take a few more weeks before it is ready.

plieven:
> Merged, thanks.
> 
> If you are interested, I am also working on finally add multithread support to libiscsi. The bare bones is in the multithreading branch right now. It still needs to protect all relevant variables/structure fields that could be accessed concurrently but it is work in progress,
> 
> If you have ideas or specific requirements you would like to see, let me know.
> 
> This is very similar/based on the multithreading in libnfs and will take a few more weeks before it is ready.

That sounds interesting. Looks like I am finally back into dealing with iscsi targets. After years of abstinence a new job finally brings me to the point where I have to integrate enterprise storages that have been designed to work well with and only with vmware to run smoothly with qemu/kvm.

PR #447, opened 2025-05-29
Fix CHAP_C Encoding and CHAP_R Comparison in CHAP Authentication

lishiao144:
#### 🐞 **Bug 1: Incorrect handling of CHAP\_C encoding during CHAP\_R computation**

In the function compute_chap_r_md5(), the chap_c parameter is expected to be a hex-encoded ASCII string (e.g., "0x1234abcd..."), which is parsed back into raw binary before being hashed into the CHAP\_R response.

However, the code that constructs CHAP_C was passing raw binary data (target_chap_c) directly into compute_chap_r_md5(), without converting it to the expected ASCII hex string format. This led to incorrect CHAP\_R generation and authentication failure.

✅ **Fix**: Before sending CHAP_C and computing CHAP_R, we now explicitly convert each byte of target_chap_c to a two-character hex string and store it in a new buffer (target_chap_c_hex), which is passed to compute_chap_r_md5().

---

#### 🐞 **Bug 2: Signed vs. unsigned byte mismatch in CHAP\_R comparison**

During the comparison of the received CHAP\_R response:

```c
if (c != iscsi->target_chap_r[i]) { ... }
```

c is unsigned char, while iscsi->target_chap_r[i] was originally declared as char. This may cause incorrect mismatches due to sign-extension when char is implicitly promoted to int during comparison (especially for byte values ≥ 128).

✅ **Fix**: The target_chap_r buffer has been updated to use unsigned char to avoid sign-extension issues during comparison with decoded CHAP_R bytes.

---

### DIFF
--- a/include/iscsi-private.h
+++ b/include/iscsi-private.h
@@ -124,7 +124,7 @@ struct iscsi_context {
 	char target_user[MAX_STRING_SIZE+1];
 	char target_passwd[MAX_STRING_SIZE+1];
 	int target_chap_i;
-	char target_chap_r[MAX_CHAP_R_SIZE];
+	unsigned char target_chap_r[MAX_CHAP_R_SIZE];
 
 	char error_string[MAX_STRING_SIZE+1];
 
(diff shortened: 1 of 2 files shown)


sahlberg:
Merged. Thanks

PR #452, opened 2025-07-08
fall-through in maintenancein unmarshal switch

glagolig:
Fixing a potential fall-through in switch in data unmarshal function for maintenance-in commands. The fall-through never happens for valid commands, but the build tools may complain about it.


sahlberg:
Merged. Thanks!

PR #454, opened 2025-07-16
examples/iscsi-dd.c: add HAVE_CLOCK_GETTIME guards

sevan:
Only show time stats if clock_gettime(3) is available. It's actually possible to have clock_gettime(3) and not yet support CLOCK_MONOTONIC on legacy systems, however on macOS, they all arrived together in macOS 10.12.
Rather than add new configure checks for CLOCK_MONOTONIC, I opted to use the existing HAVE_CLOCK_GETTIME.
libiscsi 1.20.2 now builds on Mac OS X 10.4 with examples.

### DIFF
--- a/examples/iscsi-dd.c
+++ b/examples/iscsi-dd.c
@@ -557,6 +557,7 @@ static void usage_exit(int status)
 	exit(status);
 }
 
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged. Thanks

PR #455, opened 2025-07-24
bit field order in TPG descriptor byte0

glagolig:
Re-arranging bit fields in scsi_target_port_group (representing unmarshalled data from RTPG commands) so that they match bits on the wire.
Also fixing a typo in enum scsi_alua_state.

Tested with iscsi-rtpg utility against a target with some groups having ALUA states other than "active-optimized".
Before the change all states showed up as "active-optimized" (0). After the change the states showed as expected.


sahlberg:
Merged. Thanks!

Issue #443, opened 2025-05-09
`error: 'syscall' is deprecated` on macOS (1.20.2)


misuzu:
macOS 15.4.1 24E263 arm64

```
building '/nix/store/4lb0mg0y4is2nvbj9xbqvfa227jkp7hj-libiscsi-1.20.2.drv'...
Running phase: unpackPhase
unpacking source archive /nix/store/nrbj29f92zk6f6z2df01hk5wc43xzp5r-source
source root is source
Running phase: patchPhase
Running phase: autoreconfPhase
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
aclocal: warning: couldn't open directory 'm4': No such file or directory
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
autoreconf: configure.ac: not using Intltool
autoreconf: configure.ac: not using Gtkdoc
autoreconf: running: aclocal --force -I m4
autoreconf: running: /nix/store/al2a0kqslc7s5wqcq5rybv1qnbi0kxnj-autoconf-2.72/bin/autoconf --force
autoreconf: running: /nix/store/al2a0kqslc7s5wqcq5rybv1qnbi0kxnj-autoconf-2.72/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:8: installing './ar-lib'
configure.ac:8: installing './compile'
configure.ac:9: installing './config.guess'
configure.ac:9: installing './config.sub'
configure.ac:6: installing './install-sh'
configure.ac:6: installing './missing'
examples/Makefile.am: installing './depcomp'
autoreconf: './install-sh' is updated
autoreconf: Leaving directory '.'
Running phase: updateAutotoolsGnuConfigScriptsPhase
Updating Autotools / GNU config script to a newer upstream version: ./config.sub
[...]

PR #456, opened 2025-08-01
Remove deprecated use of syscall() on Darwin

jonhermansen:
I took a stab at fixing #443. This fixes the warning about deprecated use of syscall() on macOS.

I did some very cursory testing, but I didn't specifically exercise the new mulithreading functionality. I will try to do some more extensive testing soon.

### DIFF
--- a/lib/multithreading.c
+++ b/lib/multithreading.c
@@ -39,6 +39,7 @@
 #include <sys/time.h>
 #endif
 
+#include <pthread.h>
 #include <string.h>
 #include "iscsi.h"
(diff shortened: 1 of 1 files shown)


jonhermansen:
Hmm. I should have looked a little more closely at the problem. I wonder why HAVE_PTHREAD_THREADID_NP is not defined on macOS. It may be something related to the Nix package for libiscsi.

I'll leave the PR in draft for now.

brad0:
You can close this PR.

PR #460, opened 2025-09-11
feat: read and validate unit serial number after login

plieven:
this patch adds a read of VPD page 0x80 (unit serial number) after successful login.

The serial is then validated on secutive reconnects to avoid the accidental mismatch of LUN ids if some kind of remapping appears between loss of connection and a later reconnect.

An additional url parameter force_usn is added to enforce the usn right from the beginning. If not set via url or the new iscsi_set_unit_serial_number function the usn is learned at the first successful login.

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -101,15 +101,20 @@ iscsi_inquiry_page_0x80_cb(struct iscsi_context *iscsi, int status,
 
 	if (!status) {
 		inq = scsi_datain_unmarshall(task);
-		if (!iscsi->unit_serial_number[0]) {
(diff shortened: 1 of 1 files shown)


plieven:
@sahlberg from what I found it might be that we can skip the whole discussion. for cd-rom and tape drive vpd page 0x80 might return the serial of the drive not the media.

plieven:
other sources state that some cdroms do not implement 0x80. it might be an option to check the unit serial just for direct access devices

plieven:
i would do it like in #463

PR #463, opened 2025-09-22
fix: do only query unit serial for non-removable direct access devices

plieven:
other device types might not implement evpd page 0x80 or change their serial.


sahlberg:
Merged, thanks!

